PR #24, opened 2025-07-14
intergrated permissions page with backend API, modified nginx.conf to…

susey97:
… let developers working with frontend from nginx port80, API needed for approvers modified in progress


quanwangniuniu (approved):
Good job

PR #26, opened 2025-07-14
Smp 24 login 03 implement registration api with email verification

zhuhe97:


quanwangniuniu (changes requested):
remove the migrations files

PR #33, opened 2025-07-19
Smp 31 profile 03 implement user preferences and settings api

CindyCUI423:
This PR implements User Preferences and Settings API.
Changes include:
- Created user_preferences app: models, serializers, views, urls
- Added middleware user_locale
- Added tests: test_models, test_serializers, test_views
- Added path('users/', include('user_preferences.urls')) to main urls.py
- Added language choices in settings

Passed user_preferences test:
image

Passed other tests:
image


quanwangniuniu (changes requested):
Please clean the work tree and get rid of all the chinese comments
Then PR thank you

PR #37, opened 2025-07-20
implement auth/me route

zhuhe97:
no models changed
add one endpoint named auth/me
add test_me.py
test all passed


quanwangniuniu (approved):
Nice work!

PR #48, opened 2025-07-26
changed model of access_control, set four modules allow admin to conf…

susey97:
…igure approvers from all of the users and connected with frontend page


quanwangniuniu (changes requested):
1. Get rid of chinese comments
2. Submit of the screenshots for all the test passing(run ok)

PR #56, opened 2025-07-30
let org field in role be null and assign role when users register by …

zhuhe97:
change role model in core 
update views in authentication
generate new migration files
all tests pass


quanwangniuniu (approved):
Cool stuff

PR #59, opened 2025-08-01
add two unit test for ui components related to permissions

susey97:


quanwangniuniu (changes requested):
.sh file doesn't need to be submit

PR #64, opened 2025-08-08
Smp 53 be1 03 implement asset review service tdd

zhuhe97:
1. add asset app
2. update requirements and setting
3. update the app configuration
4. Fix migrations in core and budget
5. Update CI workflow


quanwangniuniu (approved):
NIce work!

PR #68, opened 2025-08-11
Update asset review open api spec

zhuhe97:
remove url setting for tests and campaign app in backend/urls
update asset_review openAPI spec


quanwangniuniu (approved):
Nice work

PR #73, opened 2025-08-16
removed organization setting in register page and verify users immedi…

susey97:
…ately after registion


quanwangniuniu (approved):
nice work

PR #80, opened 2025-09-04
implement metric_upload app

zhuhe97:
Add an app named 'metric_upload'
Edit nginx.conf to adjust uploading file size limit
Add migration file
Edit settings.py in /backend/
Add all tests in /metric_upload/tests


quanwangniuniu (approved):
nice work

PR #113, opened 2025-10-17
SMP-237-KLAYVIYO-Model-Design-01

nexiblee:
- Added new Django app: klaviyo_draft
- Implemented models, migrations, and admin registration
- Added OpenAPI spec file for Klaviyo draft
- Jira: SMP-237


quanwangniuniu (changes requested):
1. dont commit .idea files
2. you shouldnot change docker-compose .yml or docker-compose .yml
3. you should not add anything except models

PR #117, opened 2025-10-19
Smp 229 mailchimp logic implementation 03

yswnqc:
implement in serializers.py, urls.py, views.py


quanwangniuniu (changes requested):
1. get rid of chinese comments
2. backend should not involve any frontend html tags

yswnqc:
Has modified as required, please have a review.

quanwangniuniu (approved):
Good job

PR #122, opened 2025-10-23
Smp 209 tiktok be 01 data modeling initial scaffold for tik tok drafts

zhuhe97:
1. Updated docker-compose.dev.yml and docker-compose.yml to reference different Dockerfile configurations 
2. Added installation of MediaInfo in Dockerfile
3. Added Pillow dependency in requirements.py
4. TikTok App--Added initial implementation of upload, listing, and detail endpoints for creatives (image/video) and; Added unit tests.


quanwangniuniu (changes requested):
Some of the stuff is beyond the scope of Ticket1

quanwangniuniu (approved):
Good job

PR #140, opened 2025-11-12
Tiktok add preview expiration fix fe

zhuhe97:
backend:
Set 7-day expiration when creating preview shares;
Add daily cleanup task scheduled at 02:00 UTC;
Update tests to cover expiration functionality;
frontend:
Use host local time as ad and ad group's names;
Use home page as shareable preview's background; 
Keep ad selection in memory

All tests passed


quanwangniuniu (approved):
Well done

PR #147, opened 2025-11-25
Devops test husky 20251124

Jamie-Maintainer:
Add backend test coverage and automation check in pre-push hook


quanwangniuniu (approved):
Good job

PR #185, opened 2025-12-12
Implement Migration Automation Check in CI

Jamie-Maintainer:
Implement Migration Automation Check in CI

### DIFF
--- a/.github/workflows/mediajira-ci.yml
+++ b/.github/workflows/mediajira-ci.yml
@@ -222,6 +222,17 @@ jobs:
             exit 1
           fi
 
+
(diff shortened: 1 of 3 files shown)


quanwangniuniu (changes requested):
Hi Jamie we are using postgresql, not SQLite.
Please fix the logic and then let me know please

Jamie-Maintainer:
@quanwangniuniu Because I the jira ticker mentioned "Include local sqlite or in-memory DB" then I thought we need to use sqlite3. Anyway I just replaced it with postgres, please check it out.

quanwangniuniu (approved):
Good job

PR #186, opened 2025-12-13
Smp 339 devop os 033 implement structured logging for next.js django

Jamie-Maintainer:
Implement structured logging for next.js django


quanwangniuniu (changes requested):
Fix the first issue, the logger should be setting up on docker or global settings. It should not been written manually in views file and now it is too messy. The current backend log is dangerous for external people to check and may get sensentive data leaked. So remove all the backend manually logs. For more clear scope of backend logs, it should be send to somewhere as a bundle. Same for the frontend

quanwangniuniu (approved):
Good job

quanwangniuniu (commented):
Good job

PR #187, opened 2025-12-13
BGF-02-APPROVER-issue

Xiaoboht:
Fix approver selection and display for task creation

- Introduce a `mapTaskTypeToModule` helper to translate task types (`budget`, `asset`, `retrospective`, `report`) into the backend `ModuleApprover.module` keys (`BUDGET`, `ASSET`, `CAMPAIGN`, `REPORTING`), so approvers are loaded from the correct module.
- Update `NewTaskForm` to fetch approvers by mapped module and fall back to the full user list when no module approvers are configured, preventing the "Assign an approver" dropdown from being empty.
- Simplify the approver dropdown label in `NewTaskForm` to show only the username (with email/ID used only as a fallback), improving readability for end users.
- Align `TaskDetail`’s “Next Approver” picker with the same module-based approver loading logic and fallback to all users, keeping the review flow consistent with task creation.
- Adjust the integration test `TaskCreationFlow.test.tsx` to reflect the new approver label and to assert against the username-based display.


Xiaoboht:
Refine approver selection to be project-based for task creation

What was changed
- Updated `NewTaskForm` so that the "Assign an approver" dropdown is populated exclusively from members of the selected project, instead of using module-based approvers or a global user fallback.
- Added `ProjectAPI.getProjectMembers(projectId)` to call `/api/core/projects/{project_id}/members/`, with support for both paginated (`{ results: [...] }`) and non-paginated array responses.
- Adjusted `TaskCreationFlow.test.tsx` to:
  - Mock `ProjectAPI.getProjectMembers` so approver options come from project members.
  - Mock `useProjects` to return an active project (id=1), matching the project selected in the form.
  - Update expectations to match the simplified approver label (username only) and current error message ("Network error").
 
- Simplify the approver dropdown label in NewTaskForm to show only the username (with email/ID used only as a fallback), improving readability for end users.

Behavioral impact
- When creating a task, approvers are now restricted to users who are members of the selected project.
- The approver dropdown will no longer fall back to "all users" if module approvers are missing.
- The integration test for the budget task creation flow now reflects the project-based approver logic and passes successfully.

Xiaoboht:
I added a focused unit test for NewTaskForm (NewTaskForm.approver.test.tsx) that covers the project-based approver logic:

Verifies that no approvers are loaded when no project is selected.
Verifies that, once a project is selected, getProjectMembers(project_id) is called and the project members appear as approver options.
Verifies that an empty member list shows “No approvers found”.
Verifies that selecting an approver updates current_approver_id in the parent task data.

Xiaoboht:
Backend approval validation added:

- Updated `TaskSerializer` to enforce that `current_approver_id` belongs to the same project as the task (via `ProjectMember`). If the approver is not a project member, task creation/update now returns a validation error.
- Added tests in `backend/task/tests/test_views.py`:
  - `test_create_task_with_project_member_approver_success` (task can be created when the approver is a project member)
  - `test_create_task_with_non_member_approver_fails` (task creation fails when the approver is not a project member)
- Coverage for the approval logic in `task/serializers.py` is above 80%.

quanwangniuniu (approved):
Good job xiaobo

PR #190, opened 2025-12-14
Fix internal port connection & cut sonarqube build time

Jamie-Maintainer:
Fix internal port connection & cut sonarqube build time


quanwangniuniu (approved):
Good work

PR #191, opened 2025-12-14
Allow editing current approver in Task Detail using project members

Xiaoboht:
Summary

-Updated TaskDetail.tsx so the “Current Approver” field is now a dropdown instead of read‑only text.
-The dropdown options are loaded from ProjectAPI.getProjectMembers(task.project.id), ensuring that approvers always come from the same project as the task.
-Changing the current approver sends a PATCH request via TaskAPI.updateTask to update current_approver_id on the backend, and the updated task is stored in the global task store so cards and details stay in sync.

Motivation
-Previously, users could only set an approver when creating the task. After creation, the “Current Approver” in Task Detail was read-only, so there was no way to reassign the task to another project member. This change lets users correct or update the current approver directly from the Task Detail view while still enforcing project‑scoped approvers.

Testing

Manually verified in the UI:
-Created a task with an initial approver.
-Opened Task Detail and changed “Current Approver” to another member of the same project.
-Confirmed the change is reflected in the Task Detail sidebar and in the task card list.
Frontend automated tests:
npm test -- TaskCreationFlow.test.tsx
npm test -- NewTaskForm.approver.test.tsx

### DIFF
--- a/frontend/src/components/tasks/TaskDetail.tsx
+++ b/frontend/src/components/tasks/TaskDetail.tsx
@@ -10,7 +10,7 @@ import {
 import { ScrollArea } from "../ui/scroll-area";
 import { TaskData } from "@/types/task";
 import { RemovablePicker } from "../ui/RemovablePicker";
-import { approverApi } from "@/lib/api/approverApi";
+import { ProjectAPI } from "@/lib/api/projectApi";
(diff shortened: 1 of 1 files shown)


quanwangniuniu (approved):
Good job

PR #195, opened 2025-12-19
Bgf 05 comments

Xiaoboht:
- Feature changes (TaskComment model + API + unified task-level comments UI)
- 
- Added a new TaskComment model and migration so all task types (budget/asset/retrospective/report/execution) share a single task-level comments table.
- Exposed REST endpoints GET/POST /api/tasks/<task_id>/comments/ (with project membership checks) via TaskCommentListView and TaskCommentSerializer.
- Updated frontend types and TaskAPI to support getComments / createComment, and added a unified “Comments” section to both the task detail modal and /tasks/[taskId] page so all task types can view and add comments in the same way.
- Behavior changes (Asset-specific comments hidden in task list/detail)
- 
- Updated AssetDetail to support a hideComments flag and now pass hideComments={true} in task-related contexts.
- As a result, Asset’s legacy, asset-specific comments UI is no longer rendered on the task list cards or on task detail; only the new task-level comments are shown there.
- Tests and coverage (test_task_comments + coverage)
- 
- Added task/tests/test_task_comments.py to cover listing/creating task comments, auth, project membership, and error cases (404/403).
- Verified coverage for the task app using manage.py test task and coverage tooling so that the new comment logic is exercised by tests.
- Infrastructure / migration fixes (asset/campaign/core dependencies and index rename, fixing test DB task_task issue)
- 
- Updated asset initial migration to depend on task initial migration so the task_task table is always created before assets when building databases (especially test DBs).
- Added initial migrations for the campaign app that depend on task and core migrations to avoid foreign key ordering issues, and added a core migration to rename ProjectInvitation indexes to match current Django naming.
- These migration changes resolve the previous relation "task_task" does not exist error when creating the test database and make the schema consistent and repeatable in both dev and test environments.


quanwangniuniu (changes requested):
Sort of questions before merge:
1. why we hagve to have campaign migrations
2. why we have to rename_core_project

Xiaoboht:
Q1: why we have to have campaign migrations
Because campaign already has models with a FK to task.Task but had no migrations. Django treated it as an unmigrated app and created its tables via sync_apps, sometimes before task_task existed, which caused relation "task_task" does not exist when creating the test DB. Adding a proper initial migration for campaign (with a dependency on task) puts it into the normal migration flow and guarantees task_task is created first.

Q2: why we have to rename_core_project
The core rename migration aligns the actual index names on ProjectInvitation with what Django now expects in its migration state. Without this RenameIndex migration, the DB and migration history drift apart, and makemigrations keeps generating new “rename” migrations. Keeping it once normalizes the schema and stops repeated index rename noise.

These migration changes resolve the previous relation "task_task" does not exist error when creating the test database and make the schema consistent and repeatable in both dev and test environments.

Xiaoboht:
And below is the detailed answer:
Q1: why we have to have campaign migrations
We need migrations for the campaign app because it already has database models (e.g. CampaignTask) that depend on task.Task, but previously it had no migration history. In that situation Django treats campaign as an “unmigrated app” and uses sync_apps to create its tables when building the test database. That creation order is not explicit, so in our case it tried to create campaign_task (with a FK to task_task) before task_task existed, which is what caused the relation "task_task" does not exist error during manage.py test.

By adding a proper initial migration for campaign and declaring explicit dependencies on core and task migrations, we bring campaign into the normal migrations system. That gives us a deterministic creation order (task tables first, then campaign tables), fixes the test DB failure, and ensures that all environments (local, CI, prod) can build the same schema from migrations instead of partly relying on syncdb behavior.

Q2: why we have to rename_core_project

The rename_core_project migration is there to keep the database schema in sync with what Django now expects for the ProjectInvitation model’s indexes. At some point upstream, the naming pattern for those indexes changed (e.g. from core_projec_email_abc123_idx to core_projec_email_45393c_idx), and when we ran makemigrations, Django detected that the DB index names no longer matched the names in the migration state.

Django’s way to record that change is to generate a RenameIndex migration. If we don’t keep this migration, two problems appear:

New environments (or CI) running migrate won’t have their index names aligned with the current migration history.
Future makemigrations will keep seeing the mismatch and generating more “rename” migrations, causing noise and potential divergence between environments.
By keeping this rename_core_project migration, we normalize the index names once, make the schema consistent everywhere, and give campaign’s initial migration a stable core dependency to target.

quanwangniuniu (approved):
This makes more sense and good job

PR #196, opened 2025-12-20
Implement local Kafka

Jamie-Maintainer:
Implement local Kafka

### DIFF
--- a/backend/kafka_test.py
+++ b/backend/kafka_test.py
@@ -1,0 +1,12 @@
+# kafka_test.py
+from kafka import KafkaProducer, KafkaConsumer
+
+producer = KafkaProducer(bootstrap_servers="${KAFKA_BROKER}")
+producer.send('test-topic', b'Hello from Django!')
+producer.flush()
+print("Message sent from Django")
+
+consumer = KafkaConsumer('test-topic', bootstrap_servers="${KAFKA_BROKER}")
+for msg in consumer:
+    print("Received:", msg.value.decode('utf-8'))
+    break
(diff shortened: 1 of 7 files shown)


quanwangniuniu (changes requested):
This setup is a good functional starting point, but it is still below DevOps / platform-level expectations.
The next iteration must move from “Kafka runs” to “Kafka is a usable, standardized development platform.”

Please address the following points explicitly:

Kafka Architecture
Review Kafka’s current recommended deployment model and assess moving away from Zookeeper-based setups (e.g. KRaft mode). The goal is to avoid introducing long-term technical debt in new services.

Listeners & Connectivity
Define a proper listener strategy (e.g. separate internal vs external listeners) so Kafka can be accessed reliably from containers, the host machine, and future CI/test environments. Avoid relying on a single advertised listener.

Topic Standards
Introduce explicit topic conventions (naming, partitions, retention, cleanup policies). Topics should be pre-defined and managed, not auto-created implicitly by clients.

Client Usage Patterns
Replace ad-hoc producer/consumer examples with recommended patterns (consumer groups, offset handling strategy, retry/ack configuration, graceful shutdown). Provide guidance, not just sample scripts.

Observability
Add Kafka-level observability (broker metrics, consumer lag visibility). This can be done via standard Kafka metrics exporters and visualization tools. Kafka without monitoring is not acceptable, even for dev.

Security Readiness
Even if full security is not enabled locally, the configuration should reflect a future-ready approach (e.g. client identity, auth mechanism placeholders, ACL strategy).

The next delivery should focus on operability, standards, and future scalability, rather than additional demo scripts or minimal configuration tweaks.

PR #199, opened 2025-12-22
Smp 346 devop os 361 local kafka setup with docker

Jamie-Maintainer:
Optimisation about kafka


quanwangniuniu (approved):
Approve for now, we still need to optimize it though. Lets keep going

PR #208, opened 2025-12-29
SMP-370: redirect home buttons to login page

Lin1998Simple:
This PR updates the home page buttons to correctly redirect users to the login page.

### DIFF
--- a/frontend/src/app/page.tsx
+++ b/frontend/src/app/page.tsx
@@ -1,9 +1,16 @@
+'use client'
 import React from 'react';
 import Link from 'next/link';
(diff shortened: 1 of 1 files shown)


quanwangniuniu (changes requested):
Make all the buttons in the home page direct to login page.
Also use js to controll it, dont use Link

Lin1998Simple:
I have implemented the required changes to ensure that all the buttons on the homepage now redirect users to the login page using JavaScript (instead of the Link component). I would greatly appreciate it if you could review the updated code, and  let me know if any further adjustments are needed. Thank you for your time and assistance!

quanwangniuniu (approved):
Good job

PR #218, opened 2025-12-30
Fix the issue that when pressing the "task" button, issues will apper…

Lin1998Simple:
….( add a } in line 770)


quanwangniuniu (approved):
Easy

PR #230, opened 2026-01-03
SMP-374-BGF-31-Asset-task-fix

Lin1998Simple:
Fix the asset file issue and temporarily remove the scan status


quanwangniuniu (approved):
Good job and lets keep moving

PR #237, opened 2026-01-04
Fix Chromatic workflow: remove invalid job-level if condition

EasonQi777:
- Remove job-level if condition (secrets not available in that context)
- Rely on continue-on-error for graceful failure handling
- Workflow will run and fail gracefully if secret is missing


quanwangniuniu (approved):
moderate

PR #238, opened 2026-01-04
Feat/chromatic workflow clean

EasonQi777:
Fixes the Chromatic workflow YAML validation errors by using Chromatic CLI directly instead of the GitHub Action, which allows proper use of `working-directory` for monorepo setups.


quanwangniuniu (approved):
Good fix

PR #241, opened 2026-01-04
GE-01-BE — Google Excel Data Model & ERD Design

nora-wang404:
### 1. Data model structure
- Spreadsheet → Sheet → Row / Column → Cell hierarchy.
- Row and Column are modeled as first-class entities to support dynamic insertion and reordering in future iterations.

### 2. Cell value modeling
- Cell values are stored using a `value_type` enum combined with typed value fields (`string`, `number`, `boolean`, `formula`).

### 3. Formula vs string distinction
- A cell is treated as a formula when its input starts with `=`.
- If the user intends to store literal text that starts with `=`, the input is escaped using a leading `'` (e.g. `'=SUM`).
- Escaped values are stored as raw text in `string_value`. In the future while displaying the leading escape character can be removed for presentation, while edit mode can re-show the escaped form (Stick on Google spreadsheet style).

### 4. Soft delete & uniqueness
- All uniqueness constraints are scoped with `is_deleted = false`.

### 5. Validation strategy
- Model-level validation ensures that a cell’s row and column belong to the same sheet.
- A `validate` flag is supported in `save()` to allow skipping validation for trusted bulk operations in future (e.g. copy/paste, imports).
- For normal user-driven operations, validation remains enabled by default.

### 6. Ordering trade-off
- No default ordering is enforced on the `Cell` model to avoid implicit joins and unnecessary sorting on every query.
- Ordering is expected to be applied explicitly at query time where required (e.g. table rendering).


quanwangniuniu (approved):
Good job

PR #242, opened 2026-01-04
SMP-375-Pro List View

unipirate:
1. Task list view component (TaskListView.tsx) displays the key information of tasks in a similar Jira table list view: summary, type, status, owner, approver, and due date.
2. View switching function adds " Broad View&quot; And List View&quot; The switch button enables switching between the two views.
3. The search function has added a search bar, supporting search: task summary, task description, task ID, owner username, project name, status type. The search is available in both views.
4. The pagination function displays 20 tasks per page, including complete pagination controls (previous page, next page, page number) showing the current page range and the total number of tasks.
5. The editable function in the list view, double-clicking the following fields allows editing: task summary (summary), due date (due_date). The existing TaskAPI.updateTask endpoint supports Enter to save, Escape to cancel, and the task list is automatically refreshed after editing. 6. The sorting function sorts by creation time (the latest at the top). Use the task ID as the proxy for creation time (the larger the ID, the newer).

### DIFF
--- a/backend/automationWorkflow/urls.py
+++ b/backend/automationWorkflow/urls.py
@@ -7,7 +7,6 @@ from automationWorkflow.views import (
     WorkflowViewSet,
     WorkflowNodeViewSet,
     WorkflowConnectionViewSet,
-    WorkflowRuleViewSet,
 )
 
 app_name = "workflows"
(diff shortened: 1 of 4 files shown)


quanwangniuniu (approved):
Good job

PR #243, opened 2026-01-05
Bgf 10 task start date

Xiaoboht:
Related Jira: SMP-372
Allow users to set a task start_date when creating a new task, using the same date format and UX as due_date, with start_date displayed above due_date.

Details

Frontend:

Updated NewTaskForm (NewTaskForm.tsx) to add a Start Date field (type="date") above the existing Due Date field, bound to taskData.start_date.
Updated the task creation payload in Tasks page (page.js) to send start_date (start_date: taskData.start_date || null) along with due_date.
Ensured resetFormData also clears start_date so the form state is consistent after closing/re-opening the modal.
Backend:

Reused existing support for start_date on the Task model and TaskSerializer; no schema or serializer changes were required.
Tests:

Updated the integration test TaskCreationFlow (TaskCreationFlow.test.tsx) to assert that start_date is included in the createTask payload (currently start_date: null when not set).
Impact

Task creation now has consistent date handling with Task Detail: both start_date and due_date can be edited and are persisted via the same API model.

### DIFF
--- a/frontend/src/__tests__/integration/TaskCreationFlow.test.tsx
+++ b/frontend/src/__tests__/integration/TaskCreationFlow.test.tsx
@@ -243,6 +243,7 @@ describe('TaskCreationFlow - Budget Task', () => {
           summary: 'Test Budget Task',
           description: 'Test description',
           current_approver_id: 2,
+          start_date: null,
           due_date: null
         });
       });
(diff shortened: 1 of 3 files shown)


quanwangniuniu (approved):
Good fix

PR #245, opened 2026-01-05
Devop os 085 k6 load testing infrastructure

Jamie-Maintainer:
k6 implementation


quanwangniuniu (changes requested):
Below are the concrete issues and recommended optimizations, grouped by priority.

Critical / security

Remove the real INFLUXDB_TOKEN from env.example. Do not commit actual tokens or secrets to the repo. Instead show a placeholder and add steps to generate/copy the token (you already have notes, but remove the token value).
Confirm the InfluxDB init env var in docker-compose: DOCKER_INFLUXDB_INIT_ADMIN_TOKEN should be provided from .env (not left empty). Document that users must supply it or create the token after first run.
Ensure docker-compose healthchecks + depends_on: depends_on does not wait for service readiness. Consider healthcheck + condition: service_healthy (docker compose v2) or add a startup wait/retry when other services depend on InfluxDB/Grafana.
Functional correctness / integration

Verify the k6 InfluxDB output plugin name and flags:
After building the custom image, run k6 --help (inside the image) to see available outputs. The scripts append --out 'xk6-influxdb' — confirm the plugin actually registers with that name. If not, adjust to the actual output name (often 'influxdb' for xk6-output-influxdb) or use env vars (K6_INFLUXDB_ADDR, K6_INFLUXDB_TOKEN) as shown in many plugin docs.
Suggestion for local verification: docker run --rm k6-influxdb:latest k6 version && docker run --rm k6-influxdb:latest k6 run --help to inspect outputs.
The python scripts accept vus_max and test_duration but do not pass them to k6. Either:
Remove unused args, or
Translate them into k6 args (e.g., pass environment variables to the container and consume them in scenario files, or use k6 options overrides where possible).
The docker run in the Python scripts mounts scripts_dir to /scripts:ro and runs /scripts/scenarios/*.js — that's fine, but consider running k6 as a service in docker‑compose for easier local debugging rather than docker run from Python. Or provide docker-compose.k6.yml that includes a k6 service so users can run docker compose up k6.
Docker network default: you assume mediajira_default; make this configurable and clearly document how to obtain the network name. Scripts already have K6_DOCKER_NETWORK — document usage prominently.
Code quality / maintainability

Duplicate setup code: each scenario file has almost identical setup() logic. Extract common setup logic to k6/scripts/utils/setup.js and import it to reduce duplication and maintenance overhead.
Inconsistent URL parsing: some files parse manually (regex), others call new URL(...). Pick one approach (I recommend the manual regex or a single helper in utils to avoid relying on specific k6 JS runtime features).
Optional chaining and newer JS features: be careful with optional chaining (response.body?.substring) — verify the k6 JS runtime supports it. If targeting older k6 versions, prefer safe checks (response.body && response.body.substring(...)).
Parsing JSON: many functions do JSON.parse(response.body) without verifying content-type or body. Use a parseJSON helper (you already added one) everywhere and handle null/default values to avoid runtime exceptions.
Performance / safety of test runs

Default VU / stage values are aggressive (stress up to 200 VUs). Add caution in README and possibly a safe default (smoke) and warn users before running stress/spike.
Consider adding resource limits for Docker runs or a note to run heavy tests on dedicated machines to avoid harming the developer workstation.
Add a mechanism to abort tests on egregious failures (k6 supports --abort-on-fail CLI flag). You already mention it; consider wiring that into run_* scripts as an option.
Dockerfile / build reproducibility

Building k6 using xk6 with @latest makes the build non-deterministic. Pin versions to a specific xk6/xk6-output-influxdb tag for reproducible builds.
The builder installs git via apt — consider cleaning up build caches (you already remove apt lists) and ensure the final k6 binary has correct permissions (COPY --from=builder /k6 /usr/local/bin/k6 is fine).
Consider adding a small ENTRYPOINT/healthcheck for the built image to help validate the custom image quickly.
Grafana dashboard / queries

Verify the Flux queries and dashboard JSON with your InfluxDB schema. The query-resource uses $__interval and other Grafana template variables — ensure these map to the dashboard and datasource settings after import.
The k6-dashboard JSON import may require datasource UID replacement; include a note in README to map datasources when importing.

quanwangniuniu (approved):
Good job

PR #247, opened 2026-01-06
feat:add Inline component and Storybook stories

EasonQi777:
create Inline component and storybook stories including:
default
spacing none | XS | MD | LG | XL 
align center | end | start | baseline | stretch 
justify center | end | between | around | evenly
with wrap | without wrap
button group, navigation bar, card actions, tag list, from row, custom class and nested inlines


quanwangniuniu (approved):
Good job

PR #248, opened 2026-01-06
Change spreadsheet/models.py. Add spreadsheet.yaml

nora-wang404:
This PR defines the MVP OpenAPI specification for the Spreadsheet module.
- CRUD operation for spreadsheets and sheets.
- Read/write operation for row/column/cell.
- Uses page-based pagination for spreadsheets and sheets.
- Uses offset-based scrolling for rows and columns (offset + row_limit / column_limit) to support large, grid-style data similar to Google Sheets.
- Sheet structure is managed via a dedicated resize endpoint.
- Cell access is split into range-based reads (sparse response) and batch writes with optional auto-expand.

### DIFF
--- a/backend/spreadsheet/models.py
+++ b/backend/spreadsheet/models.py
@@ -83,7 +83,7 @@ class SheetRow(TimeStampedModel):
     )
 
     class Meta:
-        ordering = ['position', 'created_at']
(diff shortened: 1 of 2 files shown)


quanwangniuniu (approved):
Good job

PR #249, opened 2026-01-06
refactor: move Inline.stories.tsx and Stack.stories.tsx to src/stories

EasonQi777:
move stories to a separate folder instead of mix with components


quanwangniuniu (approved):
Good job

PR #251, opened 2026-01-06
BGF-11 Report task

unipirate:
Front-end removes mandatory field for template:
Remove the mandatory validation and requiredFields restriction for the report template, allowing it to be empty.
Add a dropdown list for template selection: Fetch the template list from the backend /api/reports/report-templates/. In the form, it will be selectable via a dropdown, supporting None.
Request load adjustment: When creating a report, if no template is selected, pass null (backend allows it).
Modify file:
frontend/src/lib/api/reportApi.ts: Add getTemplates
frontend/src/components/tasks/NewReportForm.tsx: Load template list and dropdown selection. Template can be empty in the payload.
frontend/src/app/tasks/page.js: Remove mandatory field validation and requiredFields constraint in the form.

### DIFF
--- a/frontend/src/app/tasks/page.js
+++ b/frontend/src/app/tasks/page.js
@@ -247,7 +247,6 @@ function TasksPageContent() {
       requiredFields: [
         "title",
         "owner_id",
-        "report_template_id",
         "slice_config.csv_file_path",
       ],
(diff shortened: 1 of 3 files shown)


quanwangniuniu (approved):
Good job

PR #257, opened 2026-01-08
SMP-390 DEC-A1 – Decision Backend Models & ERD Design

unipirate:
1. Core Design Principles: Use the FSMField of django_fsm to manage the state machine (consistent with Task Asset). The core fields are immutable after commit: Controlled through model methods/attributes rather than database constraints. Independent Module: Dependent only on core.CustomUser, not other business modules. Audit Tracking: Use DecisionStateTransition to record all state changes.
2. Data model structure： Decision (Main entity)
├── Signal (One to many) - The signal that triggers a decision ├── Option (One to many) - Candidate options
├── Review (One to many) - Review after execution
├── DecisionStateTransition (One to many) - State change Audit └── CommitRecord (One to One, Optional) - Commit confirmation record
3. Key Design Points: Immutability: Check the state in clean() and save(), and prevent modification of core fields after commit. State Machine: Manage state transitions using FSMField and @transition decorators. Audit Tracking: DecisionStateTransition records all state changes. Validation: _can_commit() ensures that all conditions are met before commit. Single Option Constraint: Option.clean() ensures that only one option is selected. Index: Add indexes to commonly queried fields.


quanwangniuniu (approved):
Good job

PR #265, opened 2026-01-10
Add merge migration for task 0013 conflicts

nora-wang404:
While starting unit tests, I noticed that the task app has conflicting Django migrations on main:
two parallel 0013_* migrations without a merge migration, which causes makemigrations to fail with multiple leaf nodes.

This PR only adds the required merge migration to resolve the migration graph.
It does not modify any existing migrations or business logic, and is needed to keep makemigrations / migrate / CI working correctly.

My unit test work continues on top of this fix.


quanwangniuniu (approved):
Good job

PR #280, opened 2026-01-13
feat: add Field Pattern and State Feedback Components

kirikocat:


quanwangniuniu (changes requested):
Wrong folder, UI trigger not work, fix it

PR #281, opened 2026-01-13
Fix nginx & workflow issue

Jamie-Maintainer:
Fix nginx & workflow issue


quanwangniuniu (changes requested):
use the old configuration, but you can seperate the stages to build, scan, test and make them run synchronously.

Jamie-Maintainer:
see next

PR #293, opened 2026-01-14
Prod release 20260111

Jamie-Maintainer:
Fix nginx, workflow issues


quanwangniuniu (changes requested):
No job is triggered in this workflow, I think it is not working. 
Lets try to fix 1 issue at a time. 
Try to fix nginx first, after that then fix the auto-deploy, then fix the cicd time.
Dont do 3 work on 1 time. Hard to review and it is not working.

quanwangniuniu (changes requested):
add comment at top to tell the developers which nginx file is for developing use or production use
Also in developing mode, except the licenses, we should not remove other things. Instead, try to keep exactly the same configuration as production one. In this case, the issues that only show up in production mode might never show during developing mode. Then there will be no points to seperate them.......

quanwangniuniu (approved):
Good job Jamie

PR #300, opened 2026-01-15
feat(priority): align Jira priority levels

EasonQi777:
change to Jira style


quanwangniuniu (changes requested):
Why the github workflows configuration file is been modified?

PR #306, opened 2026-01-15
Refine calendar event dialog and drag behavior

Xiaoboht:
Summary

Add two-step event UX:
Click event → small view card next to the event (title/time/description/calendar).
Click pencil on the card → open full Google Calendar–style edit panel.
Keep full edit panel for create/edit with title, date/time, description and calendar; create uses dropdown, edit shows read-only calendar name + color.
Position both view card and edit panel as floating panels near the clicked event/time slot (no gray background overlay).
Make Week view drag behavior more natural:
Vertical drag moves event in time.
Horizontal drag moves event across days while preserving time-of-day and duration.
Resize handle adjusts only end time with a minimum duration.
Fix time editing and backend errors:
Use local localStart/localEnd state in the dialog so changing date/time actually updates start_datetime/end_datetime.
Include calendar_id and timezone when updating.
Stop sending If-Match headers to avoid 412 ETag conflicts in normal single-user editing.
How to Test

Edit an event via the small view card + pencil, change date/time, Save → event moves correctly and persists.
Create an event from an empty slot via the full panel → appears at correct time/date.
In Week view, drag events up/down (time changes) and across days (date changes, duration stays the same) without “Failed to update event time” or backend errors.

### DIFF
--- a/frontend/src/app/calendar/page.tsx
+++ b/frontend/src/app/calendar/page.tsx
@@ -19,24 +19,24 @@ import { useCalendarSidebarData } from "@/hooks/useCalendarSidebarData";
 import { useCalendarView } from "@/hooks/useCalendarView";
 import { CalendarAPI, CalendarDTO, EventDTO } from "@/lib/api/calendarApi";
 import {
+  AlignLeft,
+  Calendar as CalendarIcon,
   CalendarDays,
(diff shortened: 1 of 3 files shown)


quanwangniuniu (approved):
Good job

PR #307, opened 2026-01-15
feat(metadata): add Jira-style labels

EasonQi777:
including:
  Label
  LabelPicker


quanwangniuniu (changes requested):
Lets rename the file with JiraLabel and Jira Picker cause it might be confused in the future with other components

PR #309, opened 2026-01-16
Fix QA-II-01: Logged-in user shown as logged-out on Home

lllLucario:
## Summary
 termination issues when logged-in users were incorrectly shown as logged out on the Home page and navigation.

## Root Cause
Auth state was handled inconsistently across Home CTAs and Navbar, and initial auth loading was not properly accounted for.

## Changes
- Unified auth state handling via a single source of truth.
- Updated Home page CTA routing to reflect authenticated vs unauthenticated states.
- Ensured Navbar correctly reflects login status.

## Verification
- Verified unauthenticated users are redirected to login.
- Verified logged-in users see correct Home CTAs and navigation state.
- Verified state remains correct after page refresh.

image
image

### DIFF
--- a/backend/backend/settings.py
+++ b/backend/backend/settings.py
@@ -284,7 +284,7 @@ from .celery import app as celery_app
 __all__ = ('celery_app',)
 
 SIMPLE_JWT = {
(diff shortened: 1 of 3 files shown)


quanwangniuniu (approved):
Good job

PR #312, opened 2026-01-16
6.2/ 6.3 add Keyboard Shortcut System and Accessibility & Focus Management

kirikocat:


quanwangniuniu (changes requested):
lets rename the folder 'a11y-focus' to something else
it is not making sense.....

quanwangniuniu (changes requested):
Wrong folder I think, lets avoid this happen again.........
The trigger fail to appear for chomatic. 
Lets try to avoid make same mistakes for more than 5 times please

quanwangniuniu (changes requested):
Still need to fix

quanwangniuniu (approved):
Good job

PR #313, opened 2026-01-16
Fix auth init race by gating initialization until persisted state rehydrates

lllLucario:
### Background
In production, logged-in users could occasionally be treated as logged out when directly landing on Home (`/`).
This did not reproduce reliably in dev.

### Root Cause
Auth initialization could run before Zustand persist finished rehydrating the token from localStorage.
As a result, `initializeAuth()` sometimes evaluated auth state without the token present.

### Fix
- Introduced a `hasHydrated` flag in the auth store.
- Gated `initializeAuth()` so it only runs after persisted state rehydration completes.

### Scope
- Changes are limited to auth store rehydration and AuthProvider initialization timing.
- No changes to API logic, Home page behavior, or routing.

### Verification
- Tested with a local production build.
- Verified that logged-in users landing directly on Home consistently see the correct authenticated UI.

### DIFF
--- a/frontend/src/components/layout/Header.tsx
+++ b/frontend/src/components/layout/Header.tsx
@@ -1,5 +1,6 @@
 // src/components/layout/Header.tsx
 import React, { useState, useRef, useEffect } from 'react';
+import { useRouter } from 'next/navigation';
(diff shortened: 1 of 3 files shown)


quanwangniuniu (approved):
Good job

PR #320, opened 2026-01-17
add issue field pattern base component

EasonQi777:
a IssueField base component including Editable / loading / error / empty states
Label-value-edit structure


quanwangniuniu (changes requested):
Rename Issue to Jira please.....

PR #327, opened 2026-01-18
MED-17: Add StatusBadge story

WhiteBigBlue:
What:
- Added StatusBadge Storybook story
- Implemented keyboard navigation behavior

Why:
- Implements the StatusBadge part of MED-17


quanwangniuniu (changes requested):
Where is the actual StatusBadge Component?
We cannot only write stories...... 
Ask Eason to clarify this please

PR #337, opened 2026-01-20
Gc 01 spreadsheet

nora-wang404:
- Resize the row and column.
- Import from A1
- Beautify UI of page spreadsheet


quanwangniuniu (approved):
Good job

PR #338, opened 2026-01-20
SMP448-Task Page -- UI Redesign and Bug Fix

Lin1998Simple:
Tasks entry now requires project selection first: opening /tasks without a project shows a full-page project picker; tasks workspace appears only after a project is selected.

Project picker UI updated to a Jira-like list: header, search bar, table-style rows with column headers; recent vs all sections.

Project pinning added: pinned section at top, pin/unpin per row, pinned projects are highlighted and stored locally.

Switch project control simplified: Summary area uses a single “Switch project” button that returns to project selection.

Tasks header now shows {Project Name} - Tasks instead of just “Tasks”.

Summary/Tasks/Board tabs are in-page toggles (no redirect to dashboard).

Board is embedded inside the Tasks page, showing dashboard-style charts and activity.

Timeline view is embedded inside the Tasks page alongside List and Broad view (no route change).

Create Task form locks the project to the current one (no project selector).

Summary panel styling improved: bubble background, card-based layout, colored chips for Status/Priority/Types.

Recent activity styling refined: single-row items with softer colors and card borders.

Lead column on project picker uses project owner

### DIFF
--- a/frontend/src/app/dashboard/page.tsx
+++ b/frontend/src/app/dashboard/page.tsx
@@ -24,6 +24,12 @@ export default function DashboardPage() {
     const parsed = Number(value);
     return Number.isFinite(parsed) && parsed > 0 ? parsed : null;
   }, [searchParams]);
+  const taskId = useMemo(() => {
+    const value = searchParams.get('task_id');
+    if (!value) return null;
+    const parsed = Number(value);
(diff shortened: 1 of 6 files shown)


quanwangniuniu (approved):
Legend!

PR #365, opened 2026-01-22
feat: add profile components and storybook

EasonQi777:
Extract Profile Page UI into Storybook

Split components for reusability:

Profile header / avatar

User information display (name, email, role)

Editable fields / forms

Reuse existing UI elements (buttons, inputs, cards)

Ensure Storybook reflects all visual states (view vs edit)


quanwangniuniu (changes requested):
Starting now, our Storybook will not only include individual small components but also feature a comprehensive component that showcases how these small pieces fit together.

This overall component Storybook will provide reviewers with a visual representation of how the small components are assembled, simplifying the code review and UI review processes. If there are any areas that require enhancements or beautification, it will be easier to identify and address those specific elements.

Let’s make this adjustment to improve our workflow and ensure clarity in our reviews.

PR #370, opened 2026-01-23
Profile new

EasonQi777:
update Profile page


quanwangniuniu (changes requested):
This component is deprecated compared with our one. We should make the new updates align with the current system and everything should remain functional

PR #374, opened 2026-01-24
fix depcheck

Jamie-Maintainer:


Jamie-Maintainer:
At first, depcheck detected that there are many unused and missing dependencies, so I changed the package.json and add depcheck ignore file as well.

PR #375, opened 2026-01-24
Gc 01 spreadsheet function calculation

nora-wang404:
Formula update for spreadsheet 
- formula calculations:+-*/
- batch update by drag to fill cross
- refs cells and recalculate logic and write to backend
- SUM/AVERAGE/MIN/MAX all support drag to fill and multiple arguments


quanwangniuniu:
@Jamie6839 Hi Jamie, the backend tests are not triggered, we have to fix this

PR #387, opened 2026-01-26
fix: proactively bypass subscription check for task creation

zhangywivy:
Temporarily commented out usage limit checks in UsageTrackingMiddleware to allow task creation as requested.


quanwangniuniu (changes requested):
Sure thing! But another related issue I spoted is here in profile page. Only admin could see the plan and could subscribe. So lets make everyone should be able to fetch the plan and able to see it please

PR #417, opened 2026-01-31
insert/delete row/column and undo support, formula update.

nora-wang404:
Row and column insert and delete functions.
Support undo and formula rewrite.


quanwangniuniu (approved):
Good job

PR #418, opened 2026-01-31
Smp 464 task modification feature

shuttle666:
Added editing functions for "client" and "scaling type" tasks.
Fixed the logical bug in the calendar.


quanwangniuniu (changes requested):
get rid of chinese comments

quanwangniuniu (approved):
Good job

PR #421, opened 2026-02-01
fix(stripe): add data migration for default subscription plans

zhangywivy:
Added a Django data migration [0002_seed_plans.py] to populate the database with default plans (Free, Pro, Ultimate).


quanwangniuniu (approved):
Good job

PR #439, opened 2026-02-08
Excel operation pattern

nora-wang404:
- Recording operation patterns for row/column insert & delete  
  - Recording formula operations  
  - Displaying saved patterns in the Pattern Panel  
  - Re-applying saved patterns to different sheets and spreadsheets  
  - Basic error handling for unsupported or invalid operations


quanwangniuniu (approved):
Good job

PR #441, opened 2026-02-09
fix(BGF-24): Auto-assign Free plan to new organizations

zhangywivy:
This PR implements automatic Free plan subscription assignment for new organizations.
- Add post_save signal to automatically create Free plan subscription
- Register signals in stripe_meta app configuration
- Prevents 'no active subscription found' error for new orgs"


quanwangniuniu (approved):
Good job

PR #443, opened 2026-02-10
ui: simplify user menu navigation and improve profile page UX

zhangywivy:
- Direct profile navigation on avatar click
- Sign Out button moved to profile page
- Fix PlansSection sticky positioning
- Add InviteMembersModal integration


quanwangniuniu (approved):
Sound!

PR #453, opened 2026-02-12
Fix: unexpected Chat WebSocket connections outside chat UI

shuttle666:
Added explicit enabled control to useChatSocket (useChatSocket.ts) to gate connection lifecycle.

Updated socket start condition from userId only to userId && enabled.

Prevented auto-reconnect when socket is disabled, so non-chat contexts no longer retry/connect unexpectedly.

Made MessagePageContent the messages-page WS owner with enabled: true (MessagePageContent.tsx).

Gated ChatWidget WS with enabled: isWidgetOpen && !isMessagePageOpen (ChatWidget.tsx) to avoid background connections.

Added REST unread fallback refresh in ChatWidget on mount/login, route change, window focus, and visibility return, with 15s throttle.

Removed duplicate unread-count fetch on sidebar mount to reduce redundant API calls (Sidebar.tsx).

### DIFF
--- a/frontend/src/components/chat/ChatWidget.tsx
+++ b/frontend/src/components/chat/ChatWidget.tsx
@@ -1,6 +1,7 @@
 'use client';
 
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { usePathname } from 'next/navigation';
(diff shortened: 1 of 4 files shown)


quanwangniuniu (approved):
Good fix

PR #459, opened 2026-02-13
feat: show user role in conversation

shuttle666:
Use existing project members API and display role badges in private/group chat views.


quanwangniuniu (approved):
Sound!

PR #463, opened 2026-02-15
MED-75 Calendar Module – Google Calendar 1:1 UI Replication in Storybook

ljia0550:
Refactored calendar page into components, including calendar toolbar, sidebar panel, main view and event panel dialog

Changed some UI to match Google Calendar UI as closely as possible

Create storybook for each component


quanwangniuniu (approved):
Good job!

PR #466, opened 2026-02-17
Smp 451 chat 09 implement batch repost and repost to other

shuttle666:
- enable forward_batch for text and attachment messages
- use file-copy strategy for forwarded attachments
- keep partial_success behavior with detailed failure reasons
- include attachments + forwarded metadata in realtime/queued payloads
- improve chat list last-message preview for attachment-only messages
- update OpenAPI and backend tests


quanwangniuniu (approved):
Good fix

PR #470, opened 2026-02-18
Excel operation pattern

nora-wang404:
- Fix the file import logic and support more stable pattern reuse.
- Change pattern merge UI and move out supported now.
- New navigation to spreadsheet and delete "home" in navigation panel.


quanwangniuniu (approved):
Good job

PR #474, opened 2026-02-19
fix(chat): fix create false-fail and add delete/leave actions

shuttle666:
- rethrow create errors so duplicate-chat recovery can run
- normalize project_id/project mapping for immediate chat list visibility
- add Delete Chat / Leave Group in chat list details
- use in-app ConfirmDialog for dangerous actions
- align ConfirmDialog corner radius style


quanwangniuniu (approved):
Good job

PR #476, opened 2026-02-20
Excel operation pattern

nora-wang404:
Speed up file import by concurrent = 4 and bulk upsert.
Fix the column insert bug.


quanwangniuniu (approved):
Good job

PR #494, opened 2026-02-26
Excel operation pattern

nora-wang404:
DC-BGF-01 Decision module Malfunction  
- all type of task creating errors are fixed.
  - wrongly create task endpoint fixed.
  - task delete endpoint fixed.
  - region endpoint fixed.


quanwangniuniu (approved):
Good job Nora!

PR #501, opened 2026-03-01
Excel operation pattern

nora-wang404:
- character style
  - character size
  - freeze head line
  - current/percent/decimal adjusting


quanwangniuniu (approved):
Good job

PR #502, opened 2026-03-02
Excel operation pattern

nora-wang404:
highlight operation is working in merged pattern
  renaming patterns applis in new sheet
  multiple apply rename pattern won't create columns with duplicate names


quanwangniuniu (approved):
Good job

PR #508, opened 2026-03-04
Med 122 playwright tests–task create and task type forms

ljia0550:
1. create authentication setup to ensure users are logged in and all tests are authenticated
2. write task create test that create an asset type task
3. write approver test that check the approver is assigned correctly after task creation
4. write task type form test to cover all task types creations


quanwangniuniu (approved):
Good fix!

PR #510, opened 2026-03-05
Excel operation pattern

nora-wang404:
fix(spreadsheet): prevent duplicate range requests during scroll
  - Added strong in-flight deduplication keyed by the exact request payload.
  - Throttled scroll-triggered range loading using requestAnimationFrame to limit execution to once per frame.


quanwangniuniu (approved):
Good job

PR #515, opened 2026-03-07
SMP-497: Modified the frontend hardcoded logic for task types.

oyoNoon:
Frontend can dynamically update new types instead of manually modify many files to avoid errors.


quanwangniuniu (approved):
Good job

PR #517, opened 2026-03-08
add playwright for auth test and change playwright.config for mockdata

EasonQi777:
modify playwright.config and access test without backend data,
add auth testing


quanwangniuniu (approved):
Good job!

PR #518, opened 2026-03-09
BGF-33: Fix login blocked when Django Admin session is active

caiguanjie1104-ops:
## Summary

Fixes a bug where frontend login returns 403 when a Django Admin session is simultaneously active in the same browser.

## Problem

`LoginView` was missing `@csrf_exempt`. When a user has an active Django Admin session, the browser holds a `sessionid` cookie, which causes Django's CSRF middleware to enforce CSRF token validation on all subsequent POST requests — including the JWT-based frontend login. Since the frontend never sends a CSRF token, login would fail with 403 Forbidden.

## Solution

Added `@method_decorator(csrf_exempt, name='dispatch')` to `LoginView` in `backend/authentication/views.py`, mirroring the existing decorator already present on `RegisterView`.

**Files changed:**
- `backend/authentication/views.py` — added `@method_decorator(csrf_exempt, name='dispatch')` to `LoginView`

## Testing

**Manual:**
- Login with active Admin session → 200 OK ✅
- Login without Admin session → 200 OK ✅
- Register with active Admin session → 200 OK ✅ (already working before fix)

**Automated:**
- All existing tests pass ✅

## Jira Ticket

BGF-33 / SMP-511

### DIFF
--- a/.cursor/rules/backend-rules.mdc
+++ b/.cursor/rules/backend-rules.mdc
@@ -24,6 +24,22 @@ alwaysApply: false
 - **Consistent errors**: raise DRF `ValidationError` with field keys where possible.
 - **Permissions first**: enforce auth/authorization early and consistently.
 
+## CSRF and JWT API views
+
+- **Always apply `@csrf_exempt` to JWT-based API views** (e.g. `LoginView`, `RegisterView`).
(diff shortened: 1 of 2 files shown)


quanwangniuniu (approved):
Very strong fix

PR #524, opened 2026-03-11
feat(task):SMP-505:Allow creating task in DRAFT and explicit “Submit for review” with approver

07Coder29:
Today creation auto-submits. Optionally allow creating in DRAFT without an approver; owner then clicks “Submit for review” and must set approver at that moment. Keeps a clear “I’m done, please review” step.


quanwangniuniu (approved):
Good job!

PR #526, opened 2026-03-12
Med 125 playwright tests decisions workbench

ljia0550:
1. add playwright tests for decisions open, create, decision tree tests
2. add logo to login/register form container and hompage header section
3. finish ui polish of decision list view block


quanwangniuniu (approved):
Good Job!

PR #534, opened 2026-03-17
add profile test

EasonQi777:
including 
- logged-in user opens /profile and sees header and tab content
- profile header shows About and Contact sections
- user can update header avatar (user icon)
- user can switch tabs (Dashboard, My Organization, Subscription
- user can rapidly switch between tabs without breaking content
- user can open profile from campaigns and reach profile
- direct navigation to /profile loads profile for authenticated user
- user can edit job title and save; value is updated
- user can edit department and save; value is updated
- job title and department support long text with special characters
- job title cannot be saved as empty (falls back to previous value)
- user can edit job title multiple times and last value wins
- cancel job title edit reverts to previous value
- cancel department edit closes inline editor
- edit About section then refresh keeps profile usable


quanwangniuniu (approved):
Good job

PR #542, opened 2026-03-20
Smp-518-bgf-36-team invitation

oyoNoon:
- Fixed the errior when inviting members to a project
- Fetch roles from backend through coreAPI instead of hardcode


quanwangniuniu (approved):
Good job

PR #550, opened 2026-03-22
Cypress tests for tasks workspace

ljia0550:
1. Added user login session for Cypress e2e tests
2. Created Cypress tests for tasks workspace
3. Added data-testid attribute to attachment, subtask and task detail components to ensure their uniqueness during tests


quanwangniuniu (approved):
Good job

PR #554, opened 2026-03-22
SMP-512-Task-Filter

oyoNoon:
Purpose: Using filter to select tasks that user want to see.
Achievement: Share a same filter in board, timeline and list-view, and fixed the problems happen in selecting project


quanwangniuniu (approved):
Good fix

PR #558, opened 2026-03-24
feat(miro):Implement behaviour improvement to tools in miro

07Coder29:
Improve the annotation tool experience by fixing pen drawing issues and adding missing usability support, including consistent line thickness between curved and straight strokes, real-time stroke rendering while drawing, tooltips on hover, a related extension bar when a tool is selected, guidance for tools that require click-and-drag interaction, and arrow style selection for the arrow tool.


quanwangniuniu (approved):
Good job

PR #560, opened 2026-03-24
Task board view UI fix

ljia0550:
1. Add horizontal scroll bar inside the task board view container
2. Set max width for Jiraboard column so the long text content won't make the column too wide, which causes the content exceeds the window
3. Remove the right most create button inside the Jiraboard as it is redundant
4. Fix JiraboardCard long description overflows card issue in Safari
5. Fix Campaign page search bar overflows page container issue


quanwangniuniu (approved):
Good fix!

PR #571, opened 2026-03-27
feat(miro):Improve miro tools behaviour and simplify miro interaction

07Coder29:
The miro interaction experience needs improvement in both selection accuracy and overall usability. The current selecting tool does not correctly prioritize items when multiple lines intersect, especially when a line is enclosed by a selection box and the blank area inside that box incorrectly takes priority over another line beneath it. The canvas should also support a faster bulk-selection workflow, where drawing a rectangular selection area selects all items within that range. In addition, an eraser tool is required, and the current canvas creation flow and interface hierarchy should be simplified to make the product feel more intuitive and easier to use, closer to the usability standard of tools like Miro.


quanwangniuniu (approved):
gOOD JOB

PR #579, opened 2026-03-29
BGF-38: fix sign up issue

YanFeng02:
fixed first sign up issue let user get into main page rather than login again


quanwangniuniu (changes requested):
Dont get rid of configuration without approval.
Lets bring the docker-compose.dev.yml back

PR #584, opened 2026-03-31
Feature/cypress calendar tests

EasonQi777:
the issue is because of global css file, therefore, I modify it, and also found some inappropriate code in storybook


quanwangniuniu (approved):
Good fix

PR #592, opened 2026-04-04
Lighthouse ci test

ljia0550:
1. Integrat lighthouse auditing in local dev and github ci
2. Create scripts for lighthouse authenticated urls
3. Create README about lighthouse
4. Create lighthouse reports for urls in local and github actions artifacts in each check
5. Fix onboarding gate unneccessary horizontal scrollbar issue
6. Fix Sidebar unfixed height issue 
7. Change e2e auth test to assert the redirected url is not login instead of assert the url is some fixed ones to ensure stability


quanwangniuniu (changes requested):
lighthouse_readme should not be pushed. Thank you Ryo

quanwangniuniu (approved):
Good job

PR #604, opened 2026-04-08
MB-09-zoom-integration

YanFeng02:
This scope covers the full lifecycle of Zoom account integration within the platform. Users are able to connect their Zoom account via OAuth authorization, as well as disconnect and reconnect as needed. Once authorized, users can create scheduled Zoom meetings directly from the platform, with created meetings returning valid join and host access links and being correctly associated with the authenticated user. The scope also includes graceful handling of authorization failures such as expired or revoked tokens, and ensuring clear error feedback is provided to users when operations fail.


quanwangniuniu (approved):
Good job

PR #622, opened 2026-04-16
Med 181 production polish

ljia0550:
1. Change redirection route to /tasks after onboarding
2. Fix a  transfer onwer bug that after transfer the owner of the previous project owner can still modify the project
3. Add modal that require user to type to confirm the transfer of the owner
4. Improve the change role process
5. Delete the unimplemented serach bar of the header


quanwangniuniu (approved):
Good job

PR #624, opened 2026-04-16
Combine 01: Combine campaign-level — unify Ad Variations, Email Drafts, Ad Drafts under Campaigns

YanFeng02:
Consolidates Ad Variations, Email Drafts (Mailchimp/Klaviyo), and Ad Drafts (Facebook/Google/TikTok) into the Campaign detail page. Sidebar now only exposes Campaigns as the entry point. Each campaign tab embeds the full platform workspace with campaign-scoped filtering. Backend adds nullable media_campaign FK to all five ad/email models with manual migrations.


quanwangniuniu (approved):
Good job

PR #627, opened 2026-04-18
Med 183 page skeleton

ljia0550:
1. Create global skeleton ui component with fade animation
2. Create skeleton for profile and agent page
3. Refine Auth and onboarding checking global UI and logic
4. Remove duplicated auth check from ProtectedRoute, which is already handled in AuthProvider
5. Add auth loading policy file for extendable page skeleton that can take the loading state from auth loading UI while keeping its logic
6. store agent view in cookie and local storage to keep the view on refresh or redirection
7. Change e2e files that depend on old global loading UI


quanwangniuniu (approved):
Good job!

PR #651, opened 2026-04-26
fix: Safari password field click area 

qiwangai:
## Summary
- Fix Safari/macOS password field collapsing/click-area issue on Sign In.
- Avoid hiding WebKit textfield decoration container; ensure autofill UI doesn't block clicks.
- Stabilize `FormInput` height/line-height and reserve space for the visibility icon.


quanwangniuniu (approved):
Good fix!

PR #670, opened 2026-05-07
Add Linear integration with bidirectional task import and sync

qiwangai:
## Summary
- Added a Linear integration module for connecting Linear with MediaJira tasks.
- Implemented importing Linear issues into MediaJira tasks.
- Implemented exporting/syncing selected MediaJira tasks back to Linear.
- Moved Linear import/export actions into the Bulk edit workflow.


quanwangniuniu (approved):
Good job

PR #678, opened 2026-05-09
feat: fixed bug when user create task for specific type and it prompt…

JasonXiaoLeon:
# **Solution**

### Report Work Type: Invalid audience_type input

The audience_type field was a free-text input, allowing users to enter invalid characters (e.g. . ,) that the backend rejected. Fixed by changing the field kind from text to select in typeFieldSchemas.ts, restricting input to the valid backend enum values: client, manager, internal_team, self, other. Additionally, added conditional validation in taskTypeConfigRegistry.ts so that when audience_type is other, audience_details becomes required before submission.

image
When audience type is a value other than 'Other'
image

When audience type is a value 'Other'
image
It won't prompt the wrong hint anymore

### Three issues were identified:

communication_type options in the frontend schema did not match the backend enum values (status_update, policy_change etc. were replaced with the correct values: budget_change, creative_approval, kpi_update, targeting_change, other).

impacted_areas was not collected from the user at all, causing the backend clean() validation to always fail. Fixed by adding the field to the schema as a select with valid options, and updating getPayload to wrap the selected value in an array.

required_actions was marked required: false in the frontend schema but the backend model has no blank=True, causing empty submissions to fail. Fixed by setting required: true.
image
After filling up the blank, it successfully create without any wrong hint.
image


quanwangniuniu (approved):
Good job

PR #684, opened 2026-05-12
Smp 541

JoJo-code520:
create customer task labels, apply them to tasks, and filter the task list/board by those labels


quanwangniuniu (changes requested):
Why only 2 ci workflows are triggered? Not correct and need to be fixed

quanwangniuniu (changes requested):
we should not change mediajira-ci.yml file cause it is out of scope.
CICD_README.md is not relevent
DockerFile should never be changed(any configuration file should never be changed)
Type_META is bad design pattern applied we should come out a better way

quanwangniuniu (changes requested):
Type_META is bad design pattern applied we should come out a better way

quanwangniuniu (approved):
Good job

PR #688, opened 2026-05-15
Meta ads improvement

ljia0550:
Frontend:
- Updated Meta Ads page to load accounts in the context of the active project.
- Rebuilt AccountPicker to fetch accounts from the new listing API.
- Added account search by name and business id.
- Added pagination for account selection.
- Added labels for account status, including "shared by" and "owned"
- Added account binding and moving to current project
- Added confirmation dialog when moving an account from another project.

Backend:
- Added shared-access helpers in `facebook_integration/access.py` for:
  - accounts owned by the current user
  - accounts shared through active project membership
  - manage/sync permission checks
- Added `GET /api/facebook_integration/ad_accounts/`
  - supports `project_id`
  - supports `search`
  - supports `page` and `page_size`
  - returns paginated account results
- Updated Facebook status API to accept `project_id` and include shared project accounts.
- Extended ad account serializer with:
  - `connected_by_current_user`
  - `can_manage`
  - `can_sync`
  - `connector_name`
- Updated Meta Ads data endpoints so project members can read shared account data.
- Restricted sync and project-linking actions to users with manage/sync permissions(facebook connector or project manager).
- Added tests for shared account visibility, permissions, project binding, sync behavior, search, and pagination.


quanwangniuniu (approved):
Good job

PR #691, opened 2026-05-18
MED-205: CSM experience group and customer configuration

YanFeng02:
- Add Experience Group and Customer Django apps with project-scoped list/create APIs.
- Add admin UI for experience groups and customers ( from project card).
- Wire navigation from project select (Experience Groups / Customers).


quanwangniuniu (approved):
Good job

PR #699, opened 2026-05-23
feat(agent): Implement new render effect for chatboax

07Coder29:
Implement progressive text rendering for AI responses so messages appear word-by-word instead of as one bulk block. Use a buffered pipeline with smart scheduling to control render speed based on incoming data rate.


quanwangniuniu (approved):
good job

PR #700, opened 2026-05-24
Feature/smp 435 notification system

Uncle-Daryl-Li:
Finished notification system and message drawer


quanwangniuniu (changes requested):
Get rid of __init__.py from alerting/tests

quanwangniuniu (approved):
Good job

PR #707, opened 2026-05-29
Smp 491 meeting lifecycle state machine

Zikki-Qing:
## Summary

Implements the SMP-491 Meeting Lifecycle State Machine: a finite-state machine governing all meeting status transitions, with a full archived write-lock enforced at every layer of the stack.

### Backend

- **State machine (`lifecycle.py`)** — defines the valid transition graph (`draft → planned → in_progress → completed → archived`) with per-transition validators:
  - `in_progress`: requires ≥ 1 participant
  - `completed`: requires non-empty objective + ≥ 1 agenda item
  - `archived`: requires all action items resolved
- **Race condition fix** — `execute_transition` wraps each transition in `transaction.atomic()` + `select_for_update()`, so concurrent requests on the same meeting are serialized; the second caller reads the already-committed state and gets a clean 400 error
- **Archived write-lock (`views.py`)** — `ArchivedMeetingGuardMixin` blocks all write methods (POST/PATCH/PUT/DELETE) on every nested endpoint (agenda items, participants, artifacts, action items, document) with a 403 response
- **Service layer guard (`services.py`)** — `assert_meeting_not_archived()` raises `PermissionDenied` inside service functions, so the lock fires independently of the HTTP layer
- **`is_archived` sync** — set atomically only via `execute_transition`; direct PATCH of `is_archived` is ignored (non-archived) or blocked with 403 (archived → unarchive attempt)

### Frontend

- **`MeetingFSMActionBar`** — renders the available transition buttons for the current status with client-side preCheck gates; shows spinner during in-flight API call; replaced with a "This meeting is archived and read-only." notice when archived
- **`MeetingTypeCombobox`** — added `disabled` prop; passes `disabled={readOnly}` from `OverviewSection` so the combobox is greyed out and non-interactive on archived meetings
- **tiptap dependencies** — added 18 `@tiptap/*` packages required by `CommentComposer`

## Summary

This PR completes **SMP-491 — Meeting Lifecycle State Machine**.

It implements a finite-state machine governing all meeting status transitions and enforces a full archived write-lock across every layer of the stack.

The implementation follows a defence-in-depth approach:
- transition logic is centralised in a single `execute_transition` entry point
- concurrent transition requests are serialised through `select_for_update()` inside `transaction.atomic()`
- write-lock is enforced at the ViewSet layer via `ArchivedMeetingGuardMixin`
- write-lock is also enforced at the service layer via `assert_meeting_not_archived()`, independently of the HTTP stack
- `is_archived` is set atomically only through the state machine, never via direct PATCH
- frontend preCheck gates mirror backend validation to give immediate user feedback before the API call

### Backend

- Added `lifecycle.py` with a full FSM transition graph (`draft → planned → in_progress → completed → archived`) and per-transition validators:
  - `in_progress`: requires ≥ 1 participant
  - `completed`: requires non-empty objective + ≥ 1 agenda item
  - `archived`: requires all action items resolved
- Fixed a race condition in `execute_transition` by wrapping transitions in `transaction.atomic()` + `select_for_update()`; the second concurrent caller reads the already-committed state and receives a clean 400 error
- Added `ArchivedMeetingGuardMixin` to `views.py` blocking all write methods (POST/PATCH/PUT/DELETE) on every nested meeting endpoint with a 403 response
- Added `assert_meeting_not_archived()` guard to service functions so the write-lock fires independently of the HTTP layer
- Synced `is_archived` atomically inside `execute_transition`; direct PATCH of `is_archived: false` on an archived meeting returns 403
- Updated `openapi.yaml` with 403 responses for archived write endpoints and `is_archived` sync behaviour

### Frontend

- Added `MeetingFSMActionBar` component rendering available transition buttons with client-side preCheck gates and a spinner during in-flight requests
- Replaced the action bar with a "This meeting is archived and read-only." notice when the meeting is in archived status
- Added `disabled` prop to `MeetingTypeCombobox`; passes `disabled={readOnly}` from `OverviewSection` so the combobox is non-interactive on archived meetings
- Added 18 `@tiptap/*` packages to `package.json` required by `CommentComposer`

## Testing

Backend tests passed locally after merging latest `origin/prod-preview`:

```bash
docker compose exec backend python manage.py test meetings
# Ran 290 tests — OK
```

## Confluence URL
https://mediajirav.atlassian.net/jira/software/projects/SMP/boards/1?jql=assignee%20%3D%20712020%3A6cbb80de-27da-4a6d-880e-6b23893753a2&selectedIssue=SMP-491


quanwangniuniu (approved):
Good job

quanwangniuniu (approved):
Good job

PR #710, opened 2026-05-31
CSM-S01-05: Organisation Management

YanFeng02:
Removed root admin concept,      
  implemented self-service organisation management. Any user can create their own org and become admin automatically.
  Replaced email-based invitations with in-app notification system. Redesigned all CSM pages with shadcn/ui. Auto-create  
  org on user signup with username-based naming. Simplified OrganisationsTab with inline name editing.


quanwangniuniu (approved):
Good fix

PR #723, opened 2026-06-09
Feat/agent anomaly confirmation

Zikki-Qing:
## Summary

- Add an **Anomaly Review & Confirmation** step after analysis and before task creation — anomalies are now interactive cards instead of read-only, so false positives no longer auto-flow into task creation
- Each anomaly supports **include/exclude**, **severity editing**, and **inline description editing**; excluded anomalies stay visible but muted (greyed + strikethrough) and are preserved for audit
- The user must explicitly click **Confirm Anomalies** before any tasks are created; confirmation is **one-way** (the card locks read-only afterward)
- **Scope note:** the codebase does not create real Decision records — the actionable downstream is task creation, which is what this gate controls (`create_decision` is a dead no-op)

### Backend
- `services.py`: `_assign_anomaly_ids()` assigns stable ids at the `_run_analysis` chokepoint (idempotent; zero-anomaly datasets auto-confirm to preserve existing behaviour). New `confirm_anomalies()` action — already-confirmed safe no-op (before validation), **atomic completeness check** (rejects unknown/missing/duplicate ids so every anomaly is explicitly included/excluded), merges edits onto full anomaly objects, persists `reviewed_anomalies` + `anomalies_confirmed`, and updates the stored analysis message for faithful restore
- `services.py` / `executors.py`: `create_tasks_from_analysis` and `CreateTasksExecutor` **gate task creation** on `anomalies_confirmed`; tasks are skipped only when anomalies existed and the user excluded all of them; included anomalies attached to `commit_context` for audit
- `serializers.py`: `confirm_anomalies` action choice + `reviewed_anomalies` field
- `views.py`: chat endpoint forwards `reviewed_anomalies`; confirmation event isn't re-persisted as a duplicate card
- No migration — confirmation state lives in the existing `analysis_result` JSONField

### Frontend
- `AnomalyCard.tsx`: now interactive — include/exclude checkbox, severity dropdown, editable description, live Alerts/Signals re-grouping on severity change, pre-confirm lock warning, and one-way lock after confirm
- `AgentChatPage.tsx`: defers downstream actions until anomalies are confirmed; `handleConfirmAnomalies` + `runPostConfirmActions` (skips tasks when none included); session-restore mapper reflects reviewed/confirmed state
- `MessageList.tsx` / `agent.ts`: thread the confirm callback + `anomaliesConfirmed`; add `AnomalyItem.id`, `ReviewedAnomaly`, `WorkflowStepState.anomaliesConfirmed`, and the `anomalies_confirmed` SSE/action types

## Test plan
- [ ] Analyze a file → anomaly cards are interactive; exclude one (greys + strikethrough), change a severity (row moves Alerts↔Signals), edit a description
- [ ] No tasks created until **Confirm Anomalies** is clicked (verify in DevTools → Network: `create_tasks` only fires after `confirm_anomalies`)
- [ ] After confirm, the card is locked read-only; reload keeps it locked and does not duplicate tasks
- [ ] Exclude **all** anomalies → confirm succeeds but **no tasks created** ("All anomalies were excluded")
- [ ] Zero-anomaly dataset → no card, tasks still created (existing behaviour preserved)
- [ ] Approval **ON** → confirm routes to the task-approval step instead of auto-creating
- [ ] Backend unit tests: `python manage.py test agent`

## Linear
[agent-16](https://linear.app/mediajira/issue/MED-160/agent-16-post-analysis-anomaly-review-and-editing)

### DIFF
--- a/backend/agent/executors.py
+++ b/backend/agent/executors.py
@@ -145,6 +145,29 @@ class CreateTasksExecutor(BaseStepExecutor):
             return StepResult(success=False, error='No analysis_result in input')
 
         try:
+            # Gate: anomalies must be reviewed and confirmed before tasks.
+            if not analysis.get('anomalies_confirmed'):
+                return StepResult(
+                    success=False,
+                    error='Anomalies must be confirmed before creating tasks.',
+                )
+
+            # All-excluded: anomalies existed but none were included -> no-op
+            # success so the workflow completes cleanly. Zero-detected-anomaly
+            # runs are NOT skipped (existing behaviour preserved).
+            had_anomalies = bool(analysis.get('anomalies'))
+            reviewed = analysis.get('reviewed_anomalies') or []
+            included_anomalies = [a for a in reviewed if a.get('included', True)]
+            if had_anomalies and not included_anomalies:
+                return StepResult(
+                    success=True,
+                    output_data=input_data,
+                    sse_events=[{
+                        'type': 'text',
+                        'content': 'All anomalies were excluded; no tasks were created.',
+                    }],
+                )
+
             tasks_data = analysis.get('recommended_tasks', [])
             if not tasks_data:
                 return StepResult(success=False, error='No recommended_tasks in analysis.')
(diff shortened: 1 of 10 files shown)


quanwangniuniu (approved):
Good job

PR #725, opened 2026-06-12
Kevinhuangtaowen/med 150 agent 11 agent scope and strictness settings

07Coder29:
Users can customize how strict and how broad the AI Agent’s involvement is via settings: e.g. “only analyze spreadsheets and create decisions” vs “Agent can also create board items, tasks, etc.” Settings control what the Agent is allowed to do and how cautious it is (e.g. confirm before acting).


quanwangniuniu (approved):
Good job

PR #726, opened 2026-06-14
MED 149/AGENT 10

Uncle-Daryl-Li:
For AGENT-10, I first completed the remaining UI work that was not finished in AGENT-9.

The Workflow section is divided into two areas: **Workflows** and **Templates**. Both sections display workflows and templates using a grid-and-card layout.

When creating a workflow, users are taken to a canvas editor where they can build their own workflow by adding steps. Each step can be customized with a name and description based on specific requirements.

Both the canvas view and the workflow card hover menu include a **Template** button. Clicking this button clones the workflow into a template. When creating a template, users can specify its use case so that other project members can better understand its purpose.

Templates can be shared across different projects and organizations. Once shared, members of the corresponding project or organization can view the template under the Project Templates or Organization Templates sections. However, only the original author is allowed to modify or delete the template.

Users can also use templates created by other members as a starting point for their own workflows and add additional steps as needed.

To trigger a workflow, users need to open the canvas view and click the **Trigger** button located in the bottom action bar. Four trigger types are supported:

* **Manual** – allows workflows to be triggered manually from the hover card.
* **Polling** – allows users to select an external data source and configure a polling interval for periodic checks.
* **Instant** – listens for internal system events and triggers workflows immediately when those events occur.
* **Schedule** – designed for workflows that need to run on a fixed schedule at less frequent intervals.


quanwangniuniu (approved):
Good job

quanwangniuniu (approved):
Approved

PR #730, opened 2026-06-15
feat(csm): add conversation workspace and quick reply templates

YanFeng02:
Implement CSM conversation workspace with real-time messaging, customer portal, queue and quick reply template management.


quanwangniuniu (approved):
Good job

PR #737, opened 2026-06-25
Fix/csm settings hub

EasonQi777:
fix code for settings and build the channel configuration feature that allows administrators to set up and manage the individual support channels available on the platform.


quanwangniuniu (changes requested):
CORS_ALLOW_HEADERS = (
    *default_headers,
    'x-organization-token',
    'x-user-role',
    'x-team-id',
    'x-project-id',
    'x-active-project',
    'x-internal-refetch',
)

this has secruity issues. We should not allow this.

quanwangniuniu (approved):
Good job

PR #738, opened 2026-06-25
feat(agent): integrate draft context into Agent chat 

Zikki-Qing:
Implements AGENT-7 (Draft → Agent): users can launch the Agent from the draft list with "Ask Agent" (auto-asks for a summary) or from the draft editor with "Open in Agent" (attaches the draft as context, user types their own first question). The attached draft context is threaded on every subsequent message.

Backend
- agent/serializers.py: accept draft_context on ChatInputSerializer.
- agent/views.py: route draft_context into the orchestrator.
- agent/services.py: new answer_draft_question() reads the draft in-process by reusing notion_editor's DraftViewSet.get_queryset() for permissions and _html_to_plain_text for block rendering, then sends the draft text to the LLM. No ORM/permission/block logic is duplicated in the Agent.

Frontend
- lib/agentLaunchContext.ts: stageDraftContext / consumeDraftPreload with an autoSend boolean; shouldAutoSendDraftPreload predicate makes the "Open in Agent doesn't auto-send" rule explicit and testable.
- AgentChatPage: consumes the preload, threads draft_context on every send, gates auto-send via shouldAutoSendDraftPreload.
- NotionDraftCard: "Ask Agent" entry (autoSend: true).
- notion/[draftId]/page.tsx: "Open in Agent" entry (autoSend: false).

Constraints
- notion_editor module unchanged (verified clean).
- Permissions enforced by reusing DraftViewSet.get_queryset() in-process; the Agent has no access logic of its own.
- Read-only: Agent never creates or updates drafts. The write direction is left as a follow-up (TODO marker in answer_draft_question).

Tests
- agent/test_draft_context.py: real-ORM tests covering own-draft read, another user's draft (LLM never called), soft-deleted and nonexistent-slug errors, and handle_message routing.
- frontend launch-context tests: 5 tests covering both entry points, autoSend gating, and context persistence.

Linear
https://linear.app/mediajira/issue/MED-146/agent-7-notion-drafts-agent-integration

### DIFF
--- a/backend/agent/serializers.py
+++ b/backend/agent/serializers.py
@@ -149,6 +149,9 @@ class ChatInputSerializer(serializers.Serializer):
     )
     approval_draft = serializers.JSONField(required=False, allow_null=True)
     calendar_context = serializers.JSONField(required=False, allow_null=True)
+    # Draft → Agent context (read-only). Format: {"draftId": <pk|slug>, ...}.
+    # The Agent reads the draft via notion_editor's existing API/permissions.
+    draft_context = serializers.JSONField(required=False, allow_null=True)
     workflow_id = serializers.UUIDField(required=False, allow_null=True)
     # User-approved column mapping for confirm_columns action.
     # Format: {original_header: canonical_name or "unknown"}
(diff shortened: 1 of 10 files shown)


quanwangniuniu (approved):
Good job

PR #741, opened 2026-06-29
fix(security): scope owner-owned endpoints to the requesting user (IDOR)

zizhenliu0427:
## Summary

Several endpoints only checked `IsAuthenticated` but returned objects without ownership
filtering, so any logged-in user could read/modify another user's data by changing the
id/slug in the URL (IDOR). This PR fixes the cases that have an existing owner field, are
verified safe, and are in scope; the rest are deferred (see **Known / deferred**).

- **klaviyo EmailDraft** — owner is now set server-side from `request.user`; a client-supplied
  `user` is ignored (was: a user could create drafts attributed to another account). Write-side
  only; read was already scoped by user, so no behaviour change for normal clients.
- **facebook_meta AdCreative** — detail (GET/PATCH/DELETE), associate-media, and share-preview
  (POST/GET/DELETE) now scope to `actor=request.user`, matching the existing list view.
  Navigation is unaffected (detail ids always come from the user's own actor-filtered list);
  public sharing is unchanged (it uses a separate public token endpoint).
- **optimization** — only the A/B **Experiment** (list + detail) and **ScalingAction** endpoints
  are scoped (by `created_by` / `performed_by`). These have **no frontend caller** and previously
  returned `.objects.all()` (every user saw everyone's), so this is a pure security fix with no UI
  impact. The team-facing **optimizations** and **scaling-plans** endpoints are **untouched** — they
  were already scoped by **project membership** (`task__project_id__in=accessible_project_ids`), and
  this PR only adds slug support to their `task_id` filter, not any new restriction.
- **Scope** — only fixed what was safe and mine to fix. Endpoints that need a model/migration,
  belong to another area, or are a product decision were left out and flagged below.

## This is security hardening only — sharing / collaboration is unaffected

This PR **only** stops a user from reaching **another user's private records by id/slug**. It does
**not** change, remove, or restrict any sharing, collaboration, or team-visibility feature:

- **facebook_meta public share links** — unchanged. Recipients view a creative through the separate
  **public token endpoint** (`/facebook_meta/preview/<token>/public/`), which has no auth and no owner
  filter (unguessable `token_urlsafe(32)`). The scoping only stops user B from generating/revoking
  share links on user A's creative — you can still freely share **your own**.
- **optimization team work** — unchanged. The project-shared `optimizations` / `scaling-plans`
  endpoints still resolve visibility by project membership (see above). Only the unused, all-visible
  `experiments` / `scaling` endpoints were locked to their owner.
- **klaviyo drafts** — no sharing concept exists (the model has only a `user` FK, no team/share
  field), and read visibility is **identical before and after** (drafts were already per-user). Only
  write-side owner spoofing was fixed.
- **spreadsheet link sharing** — not touched by this PR at all (deferred; see below).

Net effect: every legitimate "let someone else see this" path keeps working; only **cross-user access
to private data** is blocked.

## Files Changed

- `backend/klaviyo/serializers.py`, `backend/klaviyo/views.py` — make `user` read-only on create;
  always set owner from `request.user`
- `backend/klaviyo/tests/test_apis.py` — regression test: a spoofed `user` in the payload is ignored
- `backend/facebook_meta/views.py` — scope AdCreative detail / associate-media / share-preview to `actor`
- `backend/optimization/views.py` — scope Experiment (list+detail) by `created_by`, ScalingAction by `performed_by`

## Expected Result

- Authenticated users can no longer read/modify another user's klaviyo drafts (write), facebook_meta
  ad creatives, or optimization experiments/scaling actions by id/slug — cross-user access returns 404.
- Behaviour preserved for all legitimate flows; **no sharing/collaboration feature is affected** (see
  the "security hardening only" section above); no migrations; no frontend changes.

## Test Plan

- [x] klaviyo API tests — 44 passed (incl. new spoofed-owner regression test)
- [x] facebook_meta + optimization tests — 327 passed, no regression
- [x] No migrations (additive access checks only)

## Known / deferred (not in this PR)

Same IDOR class, but each needs more than a safe in-scope change:

- **spreadsheet** — no membership check (any logged-in user can read/edit/delete any project's
  spreadsheet by URL). Known, and accepted from Ray. Direction when picked up: a Google-Docs–style
  model — **default private** (only project members, via the existing `IsProjectMember`), with an
  optional **"anyone with the link"** opt-in using an **unguessable share token** (the current
  human-readable slug is guessable, so it is not a safe access key on its own). Separate piece of work.
- **klaviyo Workflow** — has no owner field, so a fix needs a new owner FK + migration (and changes
  visibility); deferred.
- **agent** workflow definitions/templates, **csm** endpoints, **google_ads** media data — same class,
  but owned by other areas / need a product decision / have no direct owner field. Flagged for follow-up.


quanwangniuniu (approved):
Good job

PR #753, opened 2026-07-10
[MED-256] Recover OAuth state when third-party cookies blocked

Joestar577:
## Summary

- add backend-signed Google OAuth state so callbacks can be validated without relying only on the Django session cookie
- preserve legacy session-state validation for browsers where the session cookie is still available
- forward Google callback `state` from the frontend to Django and add sessionStorage fallback for auth persistence
- make Slack OAuth state handling resilient when local browser state is unavailable, while keeping mismatch protection

## Testing

- `docker compose -f docker-compose.dev.yml --env-file .env exec -T backend pytest authentication/tests/test_google_oauth.py -q --no-cov` -> 18 passed
- `docker compose -f docker-compose.dev.yml --env-file .env exec -T frontend env BASE_URL=http://localhost:3000 npm run test:e2e -- e2e/auth/oauth.spec.ts --project=auth-chromium --no-deps` -> 4 passed
- `docker compose -f docker-compose.dev.yml --env-file .env exec -T frontend npm run lint` -> completed successfully with existing project warnings only
- `git diff --check` -> clean

## Manual QA

- Safari Private Window with Prevent cross-site tracking enabled was attempted.
- Real Google/Slack provider QA is blocked locally because OAuth credentials are placeholders, causing upstream provider failure before the MediaJira callback.
- Callback recovery behavior is covered by targeted pytest and Playwright cookie-blocked tests.

## Migrations

- No migrations.


quanwangniuniu (approved):
Nice work

PR #754, opened 2026-07-11
MED-254: Sync notification preference UI with server truth

Popp180:
Create and Implement custom hook to help page refetch preference from server when user change sessions.


quanwangniuniu (approved):
Good job

PR #755, opened 2026-07-13
fix/MED236-eliminate approval-chain status race in batch transitions

Joestar577:
## Summary

Fixes MED-236 by serializing concurrent approval decisions on the same task's approval chain. When two approvers (or the same user in two tabs) acted on a task at the same time, both requests read the task in `UNDER_REVIEW`, both ran the FSM transition, and the second write silently clobbered the first — leaving the task in an inconsistent state and sometimes duplicating `ApprovalRecord` rows.

`TaskViewSet.make_approval` now locks the task row and re-validates its status inside a single database transaction before applying the transition. The first request to acquire the lock wins; the second blocks, then sees the already-decided status and is rejected with `409` instead of overwriting the winner. This also required making the task status-change notification signal idempotent, since it previously relied on a process-global cache that a concurrent save could clobber.

## Backend

- `backend/task/models.py`: added `Task.lock_for_transition(pk)` classmethod that wraps `select_for_update().get(pk=pk)`, centralizing the row-locking logic used by the approval endpoint.
- `backend/task/views.py`: `TaskViewSet.make_approval` now runs inside `transaction.atomic()`, re-fetches the task via `lock_for_transition`, and re-checks `task.status` under the lock. Already-decided states (`APPROVED`/`REJECTED`/`LOCKED`/`CANCELLED`) return `409 Conflict` ("This task has already been decided by another approver."); never-review-ready states (`DRAFT`/`SUBMITTED`) keep the existing `400` behavior.
- `backend/task/signals.py`: `_cache_previous_task_status` now stashes the pre-save status/anomaly values on the model instance (`_prev_status`/`_prev_anomaly`) in addition to the module-level dict. New `_consume_prev_status`/`_consume_prev_anomaly` helpers consume the value exactly once, so `notify_task_owner_on_status_change` and `notify_on_anomaly_status_change` are thread-safe and idempotent under concurrent saves of the same task.
- `backend/task/tracking_handlers.py`: `handle_task_request` tags `make-approval` writes with `transition: 'approval_decision'` in the event metadata, so tracking records the winning transition specifically. The tracking middleware only emits for `2xx` responses, so a losing `409` request is never recorded.

## Frontend

- `frontend/src/components/tasks/detail/FSMActionBar.tsx`: `run()`'s catch block now checks `isApprovalConflict(e)` — on a `409` it shows an "already been decided" toast and calls `onMutated()` to refresh the task so a stale tab immediately reflects the winning status instead of letting the user retry against outdated state.
- `frontend/src/lib/api/taskApi.ts`: added `isApprovalConflict(error)` helper to centralize the `409` check, and documented the conflict behavior on `makeApproval`.
- `frontend/e2e/tasks/task-approver.spec.ts`: added a two-tab Playwright scenario — drives a task to `UNDER_REVIEW` via API, opens it in two tabs, approves in tab A, then asserts tab B's approve attempt shows the "already decided" message and refreshes to the decided state.

Investigated a reviewer request to add double-click prevention on the action buttons before submitting; the existing `busy` state in `FSMActionBar` already disables the whole action-button group synchronously on click and restores it once the request settles, so no additional code was needed there.

## Tests

    - `pytest task/tests/test_approval_race.py --no-cov`
    - Result: 2 passed (new concurrency test — two parallel transactions racing approve/reject; `--no-cov` used to isolate from the global coverage gate)

- Django backend regression:
    - `pytest task/tests/test_approval_record.py task/tests/test_task_transitions.py task/tests/test_notifications.py tracking/tests/test_emit_event.py --no-cov`
    - Result: 35 passed

- Django backend full views + tracking suite:
    - `pytest task/tests/test_views.py tracking/ --no-cov`
    - Result: 171 passed

- Frontend targeted unit tests:
    - `npx jest src/components/tasks/taskWorkflowMvp.test.ts`
    - Result: 13 passed

- Frontend Playwright/e2e test:
    - `npx playwright test e2e/tasks/task-approver.spec.ts`
    - Result: not run locally — no `frontend/e2e/.auth/` auth state available in this environment; needs one run in an environment with e2e login state before merge.

- Diff hygiene:
    - `git diff --check`
    - Result: clean

## QA

- Demo video and QA checklist submitted in Discord `#qa-feedback`.

## Notes

- Migrations: no. No model fields or schema changed — `select_for_update` and the instance-scoped previous-status stash are both application-layer only.
- Row locking (`SELECT ... FOR UPDATE`) requires PostgreSQL; the new concurrency test is skipped automatically when the test database is SQLite.
- Out of scope (per ticket): no approval-chain redesign.


quanwangniuniu (approved):
Approved

PR #756, opened 2026-07-13
fix(frontend): namespace Zustand persist keys with versioned suffixes

Zikki-Qing:
## Summary

This PR namespaces the Zustand `persist` storage keys for `authStore` and `projectStore`, and fixes several related bugs discovered during the audit.

### Problem

- `auth-storage` and `project-storage` were generic, unversioned keys with no namespace. Any future schema change would silently load stale state on existing users with no migration path.
- Five places in the codebase hardcoded `localStorage.getItem('auth-storage')` instead of using the shared reader utility, meaning they would silently break if the key name ever changed.
- `authPersistStorage.setItem` had an early `return` after writing to localStorage/sessionStorage, which meant the auth cookie was **never written** during normal operation — only as a fallback. SSR requests therefore never received an up-to-date token.
- `profile/page.tsx` read `state.refresh` (non-existent field) instead of `state.refreshToken` when calling the delete-account endpoint, meaning the refresh token was never blacklisted on account deletion.

### Changes

**Key namespacing & migration (`api.ts`, `authStore.ts`, `projectStore.ts`)**
- Renamed persist keys: `auth-storage` → `auth-storage-v1`, `project-storage` → `project-storage-v1`
- Exported `LEGACY_AUTH_STORAGE_KEY` for use in the one-time migration
- Added `onRehydrateStorage` migration in both stores: on first load, if the new key has no real data, state is copied from the old key so existing sessions are preserved seamlessly
- Migration is idempotent: guards against Zustand's behaviour of writing empty initial state to the new key before the callback fires (checks `token || user` content, not key existence)
- Old keys are intentionally left in place during this transition; removal is tracked as a follow-up cleanup ticket

**Cookie write fix (`api.ts`)**
- Removed early `return` statements inside `authPersistStorage.setItem` so the auth cookie is always written after localStorage/sessionStorage, not only when both storage APIs are unavailable

**Hardcoded key cleanup (5 callsites)**
- `agentApi.ts`: replaced manual `localStorage.getItem('auth-storage')` + parse block with `readPersistedAuthState()`
- `TaskAttachmentsBlock.tsx` (×3): same
- `profile/page.tsx`: same, and fixed `state.refresh` → `state.refreshToken` (security fix: refresh token was not being sent to the blacklist on account deletion)

**Test setup (`jest.setup.js`)**
- Wrapped `window.matchMedia`, `IntersectionObserver`, and `ResizeObserver` mocks in `typeof window !== 'undefined'` guards so they don't throw in Node (`@jest-environment node`) test environments

### Tests

Added two new test suites:

| File | Environment | Tests |
|---|---|---|
| `persistNamespacing.test.ts` | jsdom | 6 — key naming, migration, logout scope, idempotency, cross-tab isolation |
| `persistNamespacingSSR.test.ts` | node | 2 — SSR safety (no `window` crash on import) |

All 8 tests pass.

### Out of scope / follow-up

- `chatStore` uses an unversioned key with no migration — tracked as a follow-up ticket
- Old keys `auth-storage` / `project-storage` will be cleaned up in a separate PR once the migration period ends


quanwangniuniu (changes requested):
---

### Question 1: Race Conditions and Overwrite Hazards in Multi-Tab Sessions during Migration

> **Question:**
> In `authStore.ts` and `projectStore.ts`, we implement a custom migration inside the `onRehydrateStorage` hook. If a user has **multiple browser tabs open** during the deployment of this change:
> 1. Tab A loads first, detects the absence of `auth-storage-v1`, runs the migration, hydrates its memory state, and writes the newly migrated state to `auth-storage-v1`.
> 2. Tab B loads slightly later. However, since the migration does not delete the legacy `auth-storage` key (it is left in place for safety), and there is a potential race condition before Tab B's storage events synchronize, Tab B might trigger the migration block again.
> 3. If Tab B had a different in-memory active state before the refresh, or if both try to write back to the new versioned key, they could overwrite each other's session states.
> 
> 
> Why are we not deleting the legacy key (`auth-storage`) *immediately* upon a successful migration write within a `lock` or a synchronized transactional boundary, or using a `migration-complete` flag in `localStorage` to prevent duplicate migration runs across multiple concurrently active tabs?

---

### Question 2: Hydration Desynchronization & Flashing in Axios Interceptors and SSR

> **Question:**
> In `api.ts`, we changed `AUTH_STORAGE_KEY` to `auth-storage-v1`. However, we also have Axios request/response interceptors and cookie synchronization logic (`writeAuthCookie`) that depend on the active persisted auth state.
> During hydration, `onRehydrateStorage` runs **asynchronously** after the initial client-side render (and after Axios has initialized).
> If the application makes immediate API calls during the mount phase (e.g., in a `useEffect` or inside a Next.js client component before hydration finishes):
> 1. Axios might read from the unmigrated/empty `auth-storage-v1` key, failing to attach the `Authorization` header.
> 2. The client will trigger an unauthenticated redirect or a failed request, even though valid legacy data exists in `auth-storage`.
> 
> 
> Have we verified that initial client-side API requests are blocked or delayed until `hasHydrated` is strictly `true`, or does our Axios/SSE interceptor fall back to reading the legacy key if the new one is not yet populated?

---

### Question 3: Security Leak: Cleartext Subdomain Exposure in `writeAuthCookie`

> **Question:**
> In `api.ts`, when `persistAuthTokens` is called, it triggers `writeAuthCookie(JSON.parse(value))` to write cookies for Server-Side Rendering (SSR) support.
> Now that we have migrated the storage keys, the portal login page (`frontend/src/app/(portal)/portal/login/page.tsx`) stores its auth state under a separate namespace: `portal-auth`.
> However, if the main application’s `auth-storage-v1` and the portal’s `portal-auth` both write to the same shared root domain cookie space (e.g., `.system.local`), they can continuously stomp on each other's active JWT credentials.
> Have we restricted the `Path` or `Domain` attributes of the `ms_auth` cookie in `writeAuthCookie` to prevent token leakage and cross-subdomain credential overriding between the Portal and the main application?

---

### Question 4: Incomplete Transition: Hardcoded String Hazards in Third-Party Tools

> **Question:**
> In `frontend/scripts/lhci-puppeteer-login.cjs`, we updated the Puppeteer script to look for `auth-storage-v1`:
> ```javascript
> const raw = localStorage.getItem('auth-storage-v1');
> 
> ```
> 
> 
> While this works for the Lighthouse CI configuration, there are often other developer scripts, Cypress/Playwright E2E testing configurations, local seeds, or analytics tools that mock or verify authentication by setting/getting raw values in `localStorage`.
> Did we run a global codebase sweep for **any hardcoded references** to `'auth-storage'` (such as in E2E helpers, setup scripts, or Docker setup tasks) to ensure they are updated to `'auth-storage-v1'`, preventing false-negative testing runs or broken developer environments?

---

### Question 5: Structural Integrity of `projectStore` Migration and Empty State Collisions

> **Question:**
> In `projectStore.ts`, we guard the migration by checking if the new key already has real project data:
> ```typescript
> const s = newParsed?.state;
> if (
>   s?.activeProject !== undefined ||
>   (s?.activeProjectIds?.length ?? 0) > 0 ||
>   ...
> ) return;
> 
> ```
> 
> 
> If a user had *no* active projects (e.g., a brand-new user or an empty account) prior to the deployment, their legacy state is legitimately empty (`activeProjectIds = []`).
> When they load the new version:
> 1. The store initializes with empty defaults.
> 2. The guard reads `activeProjectIds` as empty, meaning the guard condition is skipped.
> 3. It repeatedly tries to execute the legacy read migration block on every page reload, cluttering logs and consuming cycles.
> 
> 
> Why are we not writing a dedicated, explicit boolean metadata flag (e.g., `_migrated: true`) to the persisted state of `auth-storage-v1` and `project-storage-v1` to unambiguously determine if migration has been processed, instead of trying to infer it from user data shapes?

Zikki-Qing:
## Code Review Responses

### Q1: Race Conditions in Multi-Tab Sessions

Not a real risk. The migration is idempotent by design — the guard checks whether `auth-storage-v1` already contains a real `token` or `user` value before doing anything:

```typescript
const newParsed = JSON.parse(window.localStorage.getItem('auth-storage-v1'));
if (newParsed?.state?.token || newParsed?.state?.user) return;
```

Even if Tab A and Tab B both enter the migration block simultaneously, they read from the same immutable legacy key and produce identical output. The last write wins, but the content is the same — no state is corrupted. A localStorage-based mutex would add meaningful complexity for a zero-correctness-impact scenario. The legacy key is intentionally left in place to avoid breaking any in-flight sessions; removing it under a lock is unnecessary when the outcome is always the same data.

---

### Q2: Hydration Desynchronization & Axios Interceptors

Partially valid, but mitigated. `AuthProvider.tsx` already gates `initializeAuth()` behind `hasHydrated`:

```typescript
if (!hasHydrated) return;
```

This prevents token refresh and auth initialization from firing during the hydration window.

The Axios request interceptor calls `readPersistedAuthState()` directly, which reads from `localStorage` (`auth-storage-v1`) — not from Zustand in-memory state — so it is not blocked by hydration timing. The narrow window where `auth-storage-v1` is still empty (before migration writes into it) is the same pre-existing window that existed before this PR for any user who cleared storage. SSR requests are covered by the `ms_auth` cookie fallback. This is an accepted trade-off, not a regression introduced by this change.

---

### Q3: Security Leak in `writeAuthCookie`

Not a real issue. `writeAuthCookie()` sets `Path=/; SameSite=Lax` with **no `Domain` attribute**. Per RFC 6265, omitting `Domain` creates a host-only cookie scoped strictly to the exact origin — it is not shared with subdomains. If the main app and portal run on different subdomains (e.g. `app.system.local` vs `portal.system.local`), their cookies are completely isolated.

Additionally, the portal login page (`portal/login/page.tsx`) calls `localStorage.setItem('portal-auth', ...)` only — it never calls `writeAuthCookie()` or touches `ms_auth`. There is no shared cookie space between the two auth flows.

---

### Q4: Hardcoded String Hazards — Global Sweep

Yes, a full codebase sweep was performed. **14 hardcoded `'auth-storage'` references were found and updated across 13 E2E files**, covered in commit `fix(e2e): update all Playwright helpers and specs to use auth-storage-v1`:

- `e2e/tasks/task-history.spec.ts` (3 occurrences)
- `e2e/tasks/tasks-helpers.ts`
- `e2e/tasks/task-field-lock.spec.ts`
- `e2e/tasks/task-relations.spec.ts`
- `e2e/tasks/task-delete-confirm.spec.ts`
- `e2e/tasks/task-approver.spec.ts`
- `e2e/spreadsheets/spreadsheet-helpers.ts`
- `e2e/decisions/decisions-helpers.ts`
- `e2e/budget/budget-task-flow.spec.ts`
- `e2e/calendar/calendar-helpers.ts`
- `e2e/email_draft/email-draft-helpers.ts`
- `e2e/messages/messages-helpers.ts`
- `e2e/notion/notion.spec.ts`

Plus `scripts/lhci-puppeteer-login.cjs` (3 occurrences, separate commit). No remaining hardcoded references to the old key exist in the test or tooling layer.

---

### Q5: Empty State Collisions in `projectStore` Migration

Harmless in practice, acknowledged as cosmetic debt. For a new user with no projects, the legacy `project-storage` key either:

1. **Does not exist** → `if (!raw) return` exits immediately, no log noise, no cycles wasted.
2. **Exists but is empty** → migration runs but writes the same empty defaults the store already has; no corruption, no visible effect.

The `console.warn` only fires inside the `catch` block — a clean migration (even an empty one) produces no log output.

The `_migrated: true` flag approach would be strictly cleaner and eliminate this edge case entirely. We acknowledge this as technical debt and a candidate for a follow-up cleanup ticket. It does not affect correctness or user experience and does not block this PR.

quanwangniuniu (approved):
Good job

PR #758, opened 2026-07-14
Feat/med 235 hierarchy cycle v2

FFFLAREEE:
## Summary
- Add cycle detection for task parent/subtask assignments and reject invalid moves with **422** + `task_hierarchy_cycle`.
- Wire validation through task services and API (add subtask, move subtask).
- Add **Parent picker** on subtask detail to reassign parent via slug-based move API.
- Paginated parent candidate list; retain previously selected parents after reassign.
- Backend tests, Jest tests, and Playwright E2E spec.
## Product note
Only **one hierarchy level** (parent → subtask). Normal UI cannot trigger a real cycle; **422** is defensive for dirty data / direct API.
## Test plan
- [x] `task.tests.test_hierarchy_cycle` — 11/11 OK
- [x] Jest `TaskParentPicker` + `taskHierarchyApiError` — 12/12 PASS
- [x] `npm run build` — pass
- [x] Manual QA: parent reassign → move 200, persists after refresh
- [ ] Playwright `task-parent-cycle.spec.ts` — flaky locally on detail page load
Ticket: **MED-235**


quanwangniuniu (changes requested):
### Question 1: Race Conditions in Concurrent Cyclic Assignments (DFS Validation Lock Design)

> **Question:**
> In `models.py`, the `would_create_task_hierarchy_cycle` helper uses a Depth-First Search (DFS) query-loop on `TaskHierarchy` to validate that no cycle is introduced.
> If two users concurrently attempt to cross-assign the same two tasks (e.g., User A sets Task 1 as parent of Task 2, while User B concurrently sets Task 2 as parent of Task 1), both requests could execute the `would_create_task_hierarchy_cycle` check before either transaction is committed. Both checks would read the stale state, determine "no cycle exists", and successfully write the records—resulting in a hard-locked cyclic hierarchy after commit.
> Should we introduce a database-level lock (such as `select_for_update` on the involved `Task` rows) within our validation services or models to serialize these checks and strictly guarantee database integrity under high concurrency?

---

### Question 2: Memory & API Bottlenecks with Sequential Pagination in `getAllTasks`

> **Question:**
> In `frontend/src/lib/api/taskApi.ts`, the new `getAllTasks` utility implements a sequential `do...while` loop that eagerly page-walks up to 100 pages of task data:
> 1. In standard production environments with large projects, fetching and accumulating thousands of `TaskData` entities into a single memory array can cause significant frontend heap memory pressure and UI freeze.
> 2. Initiating dozens of consecutive HTTP requests sequentially on component mount could easily trigger backend Rate-Limiting/WAF block rules, or saturate browser network queues.
> 
> 
> Can we refactor the `TaskParentPicker` to rely on a debounced, search-based autocompleter that queries only matching candidates, rather than eagerly downloading the entire project's flat task space on mount?

---

### Question 3: Django Memory Object Desynchronization upon Failure in `reassign_subtask_parent`

> **Question:**
> In `backend/task/services.py`, the `reassign_subtask_parent` function prevents the `post_delete` orphan signal handler from purging the child during a move by executing:
> ```python
> child_task.is_subtask = False
> child_task.save(update_fields=['is_subtask'])  # Step 1
> TaskHierarchy.objects.filter(...).delete()      # Step 2
> new_parent.add_subtask(child_task)              # Step 3
> 
> ```
> 
> 
> While this is wrapped in a `@transaction.atomic` block to ensure database rollbacks, if Step 3 fails (e.g., due to an unforeseen validation error or database issue), the Python in-memory `child_task` instance remains mutated with `is_subtask = False`. If this same object instance is referenced downstream or in-memory in subsequent operations during the request-response lifecycle, it will reflect a desynchronized state.
> Can we avoid modifying and saving the model instance twice? Is there a way to update the relation records directly (e.g. updating the foreign keys of the `TaskHierarchy` entry) to bypass the trigger conditions of the orphan handler without mutating the core task record's state?

---

### Question 4: Architectural Redundancy between Multi-node Cycle Detection and 1-Level Nesting Constraints

> **Question:**
> Throughout `models.py` and `services.py`, we strictly enforce a flat hierarchy:
> `A subtask cannot have subtasks. Only 1 level of nesting is allowed.`
> If a task can strictly *only* be either a root-level parent or a single-level child, a circular reference check is structurally trivial: a task that already has children cannot be assigned a parent, and a task that has a parent cannot have children.
> Why are we implementing a complex DFS traversal (`would_create_task_hierarchy_cycle` supporting N-node paths) if the 1-level nesting rules prevent anything deeper than 2 nodes? Is this DFS infrastructure written to prepare for removing the 1-level nesting limitation and transitioning to an arbitrary tree/DAG model in the near future? If so, how will we transition away from the binary `is_subtask` boolean flags?

---

### Question 5: Frontend State Reversion & Handling Errors on Failed Moves

> **Question:**
> In `TaskParentPicker.tsx`, we catch errors thrown by `TaskAPI.moveSubtask`, set `inlineError`, and trigger a toast.
> However, if the API request fails (e.g., due to a validation error, 422 cycle rejection, or 500 error), the UI component’s internal React state (`parentId`) or display value might become desynchronized with the backend, or the picker could get stuck in a state that doesn't accurately reflect reality.
> When an API rejection occurs, do we have a robust mechanism to revert the state back to the pre-change `currentParentId` instead of relying on the user to manually refresh the page?

FFFLAREEE:
> ### Question 1: Race Conditions in Concurrent Cyclic Assignments (DFS Validation Lock Design)
> > **Question:**
> > In `models.py`, the `would_create_task_hierarchy_cycle` helper uses a Depth-First Search (DFS) query-loop on `TaskHierarchy` to validate that no cycle is introduced.
> > If two users concurrently attempt to cross-assign the same two tasks (e.g., User A sets Task 1 as parent of Task 2, while User B concurrently sets Task 2 as parent of Task 1), both requests could execute the `would_create_task_hierarchy_cycle` check before either transaction is committed. Both checks would read the stale state, determine "no cycle exists", and successfully write the records—resulting in a hard-locked cyclic hierarchy after commit.
> > Should we introduce a database-level lock (such as `select_for_update` on the involved `Task` rows) within our validation services or models to serialize these checks and strictly guarantee database integrity under high concurrency?
> 
> ### Question 2: Memory & API Bottlenecks with Sequential Pagination in `getAllTasks`
> > **Question:**
> > In `frontend/src/lib/api/taskApi.ts`, the new `getAllTasks` utility implements a sequential `do...while` loop that eagerly page-walks up to 100 pages of task data:
> > 
> > 1. In standard production environments with large projects, fetching and accumulating thousands of `TaskData` entities into a single memory array can cause significant frontend heap memory pressure and UI freeze.
> > 2. Initiating dozens of consecutive HTTP requests sequentially on component mount could easily trigger backend Rate-Limiting/WAF block rules, or saturate browser network queues.
> > 
> > Can we refactor the `TaskParentPicker` to rely on a debounced, search-based autocompleter that queries only matching candidates, rather than eagerly downloading the entire project's flat task space on mount?
> 
> ### Question 3: Django Memory Object Desynchronization upon Failure in `reassign_subtask_parent`
> > **Question:**
> > In `backend/task/services.py`, the `reassign_subtask_parent` function prevents the `post_delete` orphan signal handler from purging the child during a move by executing:
> > ```python
> > child_task.is_subtask = False
> > child_task.save(update_fields=['is_subtask'])  # Step 1
> > TaskHierarchy.objects.filter(...).delete()      # Step 2
> > new_parent.add_subtask(child_task)              # Step 3
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > While this is wrapped in a `@transaction.atomic` block to ensure database rollbacks, if Step 3 fails (e.g., due to an unforeseen validation error or database issue), the Python in-memory `child_task` instance remains mutated with `is_subtask = False`. If this same object instance is referenced downstream or in-memory in subsequent operations during the request-response lifecycle, it will reflect a desynchronized state.
> > Can we avoid modifying and saving the model instance twice? Is there a way to update the relation records directly (e.g. updating the foreign keys of the `TaskHierarchy` entry) to bypass the trigger conditions of the orphan handler without mutating the core task record's state?
> 
> ### Question 4: Architectural Redundancy between Multi-node Cycle Detection and 1-Level Nesting Constraints
> > **Question:**
> > Throughout `models.py` and `services.py`, we strictly enforce a flat hierarchy:
> > `A subtask cannot have subtasks. Only 1 level of nesting is allowed.`
> > If a task can strictly _only_ be either a root-level parent or a single-level child, a circular reference check is structurally trivial: a task that already has children cannot be assigned a parent, and a task that has a parent cannot have children.
> > Why are we implementing a complex DFS traversal (`would_create_task_hierarchy_cycle` supporting N-node paths) if the 1-level nesting rules prevent anything deeper than 2 nodes? Is this DFS infrastructure written to prepare for removing the 1-level nesting limitation and transitioning to an arbitrary tree/DAG model in the near future? If so, how will we transition away from the binary `is_subtask` boolean flags?
> 
> ### Question 5: Frontend State Reversion & Handling Errors on Failed Moves
> > **Question:**
> > In `TaskParentPicker.tsx`, we catch errors thrown by `TaskAPI.moveSubtask`, set `inlineError`, and trigger a toast.
> > However, if the API request fails (e.g., due to a validation error, 422 cycle rejection, or 500 error), the UI component’s internal React state (`parentId`) or display value might become desynchronized with the backend, or the picker could get stuck in a state that doesn't accurately reflect reality.
> > When an API rejection occurs, do we have a robust mechanism to revert the state back to the pre-change `currentParentId` instead of relying on the user to manually refresh the page?


Q1 — Race Conditions in Concurrent Cyclic Assignments
Yes — we added row-level locking to serialize hierarchy writes.

We agreed this was a real race: two concurrent cross-assignments (1→2 and 2→1) could both pass DFS against stale state before either transaction committed.

Fix: _lock_tasks_for_hierarchy_write() in backend/task/services.py uses Task.objects.select_for_update().filter(pk__in=...).order_by('pk') at the start of both add_subtask_to_parent and reassign_subtask_parent. Involved tasks are locked in ascending PK order (consistent lock ordering reduces deadlock risk). Validation and writes then run against a consistent snapshot.

Tests: backend/task/tests/test_hierarchy_race.py (PostgreSQL only, same pattern as test_approval_race.py).

The second concurrent request blocks until the first commits; its DFS check then sees the updated hierarchy and correctly rejects the cycle.

Q2 — Memory & API Bottlenecks with getAllTasks
Yes — TaskParentPicker no longer eagerly loads the full project task list on mount.

Before: mount triggered getAllTasks, which sequentially page-walked up to 100 pages.

After:

On open, we show the current parent from task.parent_relationship (no list API call).
Search runs only after the user types (≥1 character), debounced at 300ms.
Each search issues one request: GET /api/tasks/?search=...&has_parent=false&page_size=20&page=1.
Backend search matches summary__icontains only (not slug, to avoid false positives) and ranks exact → prefix → contains matches first.
This removes mount-time heap pressure and the burst of sequential HTTP requests. getAllTasks remains in taskApi.ts for other callers but is no longer used by the parent picker.

Q3 — Django In-Memory Desync on Failure in reassign_subtask_parent
Valid concern in theory; low practical risk in our current request path. No refactor planned in this PR.

The is_subtask = False save before deleting the old TaskHierarchy row is intentional: it prevents the orphan post_delete signal handler from deleting the child mid-move. The whole sequence is wrapped in @transaction.atomic, so if Step 3 fails, the database rolls back including the Step 1 write.

You're correct that the in-memory child_task instance can remain is_subtask=False after an exception. However, in move_subtask the view catches the error and returns immediately — it does not reuse child_task for further operations in the same request. The next request loads a fresh instance from the DB.

Alternatives considered:

Updating TaskHierarchy FK in place would avoid toggling is_subtask, but would require reworking the orphan handler / signal assumptions — a larger change with limited benefit for MED-235.
A lightweight follow-up, if desired: child_task.refresh_from_db() before re-raising inside the service. Happy to add that in a follow-up commit if you'd like the extra safety.
Q4 — DFS vs 1-Level Nesting Redundancy
The 1-level rules and DFS serve different purposes; DFS is not redundant.

The flat-hierarchy constraints (parent cannot be a subtask, subtask cannot have children) prevent deep trees, but they do not fully replace cycle detection:

Concurrent cross-assign (Q1): without locking + cycle check, two root tasks can still end up in a 2-node cycle after both writes commit — nesting rules alone don't serialize that.
Reassign with excluding_edges: when moving B from A→C, we temporarily ignore the A→B edge while validating C→B. That requires graph reachability logic, not just local 1-level checks.
Defense in depth: nesting rules catch common invalid cases early with clear error messages; DFS is the authoritative graph integrity check.
Future multi-level nesting: yes, the DFS infrastructure is intentionally forward-compatible. If we lift the 1-level limit, the cycle validator stays; we'd gradually deprecate the denormalized is_subtask boolean in favor of deriving role from TaskHierarchy (parent/child queries). That migration is out of scope for MED-235, but the current design doesn't block it.

Q5 — Frontend State Reversion on Failed Moves
Yes — failed moves revert UI state; no manual refresh needed.

In TaskParentPicker.tsx:

We do not optimistically update parentId before the API call.
setParentId(nextParentId) runs only after a successful TaskAPI.moveSubtask.
On failure (422 cycle, 400 validation, 500, etc.), the catch block resets parentId to currentParentId (derived from task.parent_relationship).
Hierarchy cycle errors (422) show an inline error; other errors show a toast.
The picker trigger label is derived from parentId via selectedParent, so display stays aligned with the last known good backend state when a move is rejected.

quanwangniuniu (changes requested):
Thanks for adding cycle detection — this is an important safety net. A couple of follow-ups/questions and a requested change:

1. Concurrency / race conditions
 - How does the service protect against concurrent moves that could still create a cycle (two requests racing and each basing its check on the pre-move state)? Right now the check looks synchronous — but without DB-level protection a race can slip through.
 - Please add either:
   a) transactional locking (e.g., SELECT ... FOR UPDATE on the involved task rows) around the cycle-check + move, or
   b) a database-side constraint/trigger that enforces acyclicity, or
   c) an optimistic-retry pattern with a unique detect-and-retry loop plus tests demonstrating correctness under concurrency.

2. Failure contract
 - Confirm the exact error payload and fields returned for the 422/task_hierarchy_cycle case (HTTP body shape, error code string, user-facing message). Ensure it’s stable and documented so clients (UI, integrations) can react consistently.

3. Tests
 - Add automated tests simulating concurrent moves that would create a cycle (multi-threaded/unit tests or integration tests) and assert that one move succeeds and the other fails with the task_hierarchy_cycle error.

Could you implement one of (a)/(b)/(c) above and add the concurrency tests and a short note in the API docs describing the 422 response shape?


Nice work on the parent picker and paginated candidate list. A few potential problems to address:

1. Slug-based moves are brittle
 - Using only slugs to identify parent targets can be ambiguous (duplicate slugs across projects/workspaces) and fragile if slugs change. Recommend using task IDs (stable unique identifiers) in the move API or requiring explicit disambiguation (workspace + slug).

2. Pagination + retention of previously-selected parent
 - The picker says it "retains previously selected parents after reassign", but pagination can hide the selected parent on subsequent loads. Please ensure the UI:
   - Always includes the currently selected parent in the candidate results (top of list) even if it’s outside the current page, or
   - Prefetch/merge the selected parent into the currently displayed page, and
   - Prevents stale selection UX (e.g., immediately re-fetch to confirm parent is still valid after move).

3. Error handling UX
 - Ensure the UI treats the 422/task_hierarchy_cycle response specially (show clear user-facing message and do not allow a silent retry). Add unit + E2E tests verifying the picker shows the right message when backend rejects a move.

Can you switch the move API to accept task IDs (or a combined unique identifier), and update the picker to include the selected parent outside of the pagination window plus tests for the 422 path?


Thanks for the Playwright spec. Two concerns and a proposed plan:

1. Flaky E2E (task-parent-cycle.spec.ts)
 - The PR note says the Playwright spec is flaky on detail page load. Flaky tests often mask real bugs (timing, network races). Please:
   - Stabilize the test using deterministic waits (waitForResponse for relevant API calls, waitForSelector for elements, avoid arbitrary timeouts).
   - Use explicit test fixtures (seeded DB state) so the test is deterministic.
   - If the UI needs some asynchronous background work before being ready, expose a ready flag in the API or a deterministic network request the test can wait on.

2. Existing DB data integrity
 - The defensive 422 is good for API-level protection, but does the repo include a migration or a one-time scrub to detect and fix any pre-existing cycles in production data? If not, add:
   - A migration/check that searches for cycles and either fixes or flags them, plus a safe rollback/action plan.
   - Telemetry/logging for rejected move attempts (counts, affected task IDs) so we can monitor for attacks or mass data issues.

Please stabilize the Playwright test with deterministic waits/fixtures, and add a migration or admin-check to find and handle existing hierarchy cycles. Also include CI steps to run the Playwright test reliably (retries or flakiness mitigation).

FFFLAREEE:
> Thanks for adding cycle detection — this is an important safety net. A couple of follow-ups/questions and a requested change:
> 
> 1. Concurrency / race conditions
> 
> * How does the service protect against concurrent moves that could still create a cycle (two requests racing and each basing its check on the pre-move state)? Right now the check looks synchronous — but without DB-level protection a race can slip through.
> * Please add either:
>   a) transactional locking (e.g., SELECT ... FOR UPDATE on the involved task rows) around the cycle-check + move, or
>   b) a database-side constraint/trigger that enforces acyclicity, or
>   c) an optimistic-retry pattern with a unique detect-and-retry loop plus tests demonstrating correctness under concurrency.
> 
> 2. Failure contract
> 
> * Confirm the exact error payload and fields returned for the 422/task_hierarchy_cycle case (HTTP body shape, error code string, user-facing message). Ensure it’s stable and documented so clients (UI, integrations) can react consistently.
> 
> 3. Tests
> 
> * Add automated tests simulating concurrent moves that would create a cycle (multi-threaded/unit tests or integration tests) and assert that one move succeeds and the other fails with the task_hierarchy_cycle error.
> 
> Could you implement one of (a)/(b)/(c) above and add the concurrency tests and a short note in the API docs describing the 422 response shape?
> 
> Nice work on the parent picker and paginated candidate list. A few potential problems to address:
> 
> 1. Slug-based moves are brittle
> 
> * Using only slugs to identify parent targets can be ambiguous (duplicate slugs across projects/workspaces) and fragile if slugs change. Recommend using task IDs (stable unique identifiers) in the move API or requiring explicit disambiguation (workspace + slug).
> 
> 2. Pagination + retention of previously-selected parent
> 
> * The picker says it "retains previously selected parents after reassign", but pagination can hide the selected parent on subsequent loads. Please ensure the UI:
>   
>   * Always includes the currently selected parent in the candidate results (top of list) even if it’s outside the current page, or
>   * Prefetch/merge the selected parent into the currently displayed page, and
>   * Prevents stale selection UX (e.g., immediately re-fetch to confirm parent is still valid after move).
> 
> 3. Error handling UX
> 
> * Ensure the UI treats the 422/task_hierarchy_cycle response specially (show clear user-facing message and do not allow a silent retry). Add unit + E2E tests verifying the picker shows the right message when backend rejects a move.
> 
> Can you switch the move API to accept task IDs (or a combined unique identifier), and update the picker to include the selected parent outside of the pagination window plus tests for the 422 path?
> 
> Thanks for the Playwright spec. Two concerns and a proposed plan:
> 
> 1. Flaky E2E (task-parent-cycle.spec.ts)
> 
> * The PR note says the Playwright spec is flaky on detail page load. Flaky tests often mask real bugs (timing, network races). Please:
>   
>   * Stabilize the test using deterministic waits (waitForResponse for relevant API calls, waitForSelector for elements, avoid arbitrary timeouts).
>   * Use explicit test fixtures (seeded DB state) so the test is deterministic.
>   * If the UI needs some asynchronous background work before being ready, expose a ready flag in the API or a deterministic network request the test can wait on.
> 
> 2. Existing DB data integrity
> 
> * The defensive 422 is good for API-level protection, but does the repo include a migration or a one-time scrub to detect and fix any pre-existing cycles in production data? If not, add:
>   
>   * A migration/check that searches for cycles and either fixes or flags them, plus a safe rollback/action plan.
>   * Telemetry/logging for rejected move attempts (counts, affected task IDs) so we can monitor for attacks or mass data issues.
> 
> Please stabilize the Playwright test with deterministic waits/fixtures, and add a migration or admin-check to find and handle existing hierarchy cycles. Also include CI steps to run the Playwright test reliably (retries or flakiness mitigation).

Thanks for the detailed review. Below is how this PR addresses each comment, along with the items deferred to follow-up work.

## Comment 1 — Concurrency, 422 Contract, and Tests

### Concurrency

Implemented **option (a): transactional row-level locking**.

Hierarchy write operations now use `_lock_tasks_for_hierarchy_write()` with `select_for_update()` in `backend/task/services.py`. The cycle check and parent move are executed within the same database transaction, preventing concurrent requests from both passing validation against stale pre-move data.

### 422 Failure Contract

The cycle rejection contract is stable and documented.

For both parent moves and add-subtask operations, the API returns:

```http
HTTP 422
```

```json
{
  "detail": "Cannot set this parent: it would create a circular task hierarchy.",
  "code": "task_hierarchy_cycle"
}
```

This contract is documented in:

* `openapi/openapi_spec/task.yaml` under `TaskHierarchyCycleErrorResponse`
* The `move_subtask` view docstring

Clients should match against HTTP `422` and/or the error code `task_hierarchy_cycle`.

### Concurrency Tests

Added PostgreSQL concurrency coverage in:

`backend/task/tests/test_hierarchy_race.py`

The test starts two threads performing conflicting hierarchy moves. One operation succeeds, while the other is rejected with `TaskHierarchyCycleError`, HTTP `422`, and the `task_hierarchy_cycle` error code.

Additional cycle coverage is included in:

`backend/task/tests/test_hierarchy_cycle.py`

This includes parent moves using both task slugs and numeric task IDs.

## Comment 2 — Parent Picker, Task IDs, Pagination, and 422 UX

### Task IDs in the Move API

The parent picker calls the move endpoint using stable task identifiers in the URL path.

The backend accepts both numeric task IDs and slugs through the task-specific `resolve_task_lookup_kwargs()` implementation in:

`backend/task/lookups.py`

This behaviour applies only to tasks; other resources remain slug-only.

The request body continues to use the numeric `old_parent_id`.

Task slugs are globally unique and immutable after creation, so cross-project slug ambiguity does not apply to the current data model.

### Pagination and Retained Parent

`TaskParentPicker` merges pinned and retained parents into the searchable result list using `rememberParent` and debounced search.

This ensures that the current or newly selected parent remains visible even when it is outside the active search-result page.

On a failed move:

* The selection reverts to the previous parent.
* The error is displayed to the user.

On a successful move:

* `onUpdated()` refreshes the task state.

### 422 UX

The frontend handles HTTP `422` and `task_hierarchy_cycle` as a dedicated error case.

For cycle errors:

* An inline error is rendered through `task-parent-picker-error`.
* The selected parent is reverted.
* The request is not silently retried.

Other errors continue to use the standard toast notification.

### Test Coverage

Frontend unit test:

`frontend/src/__tests__/components/tasks/TaskParentPicker.test.tsx`

Playwright E2E test:

`frontend/e2e/tasks/task-parent-cycle.spec.ts`

The E2E test mocks the `422` response to provide deterministic UI coverage.

The real three-node HTTP `422` contract is covered by backend tests. The public API cannot create the required invalid hierarchy chain without bypassing normal database rules because the product currently supports only one level of nesting.

## Comment 3 — E2E Stabilisation, Existing Data, and CI

### E2E Improvements Included in This PR

The following changes were made to improve E2E stability:

* Added deterministic `waitForResponse` checks for task-detail GET requests, parent search, and move POST requests.
* Added API-based test fixtures for task creation, linking, deletion, and `afterEach` cleanup.
* Removed `docker exec` database seeding.
* Covered the UI `422` state through route mocking and the real API `422` contract through backend tests.
* Updated authentication setup to use API login and `expect.poll()` instead of fixed sleeps.
* Added `E2E_USE_EXISTING_SERVER=1` support to skip the Playwright `webServer` when the application is already served through Docker and nginx.
* Added `beforeAll` route warm-up for `/tasks/[taskId]` to reduce cold-compilation `502` errors.

Local E2E execution may still fail when `frontend-dev` becomes unstable during Next.js restarts and temporarily returns `502`.

The backend and frontend unit tests remain the merge gate for the `422` contract. The Playwright specification has been verified successfully when the local stack is healthy.

### Existing Database Integrity — Deferred

This PR adds write-path validation and returns HTTP `422` when an invalid hierarchy move is attempted.

It does not include a production data scrub, migration, or structured warning logs for hierarchy cycles.

Proposed follow-up work:

* Add a `detect_task_hierarchy_cycles` management command.
* Make the command report-only by default, with `--dry-run` support.
* Add warning logs for rejected hierarchy moves.
* Log task IDs only and avoid logging secrets or sensitive data.

### CI Playwright Coverage — Deferred

Playwright has not been added to `mediajira-ci.yml` in this PR because it requires:

* A running Docker Compose application stack
* Valid E2E authentication credentials
* Additional CI environment setup

The current local Playwright configuration uses:

* `retries: 1`
* Trace collection on retry

quanwangniuniu (approved):
Approved!

PR #763, opened 2026-07-20
Add SBOM, permission matrix, data export, and OAuth state hardening

qiwangai:
## Summary
- Generate CycloneDX SBOMs for tagged releases and document release artifact locations
- Add read-only project permission matrix endpoint and admin UI with warnings
- Add GDPR personal data export requests with async ZIP assembly, signed downloads, and profile UI
- Harden OAuth callback state validation with signed, time-bound, single-use state values

## Validation
- docker compose -f docker-compose.dev.yml --env-file .env exec -T backend python manage.py check
- docker compose -f docker-compose.dev.yml --env-file .env exec -T backend pytest core/tests/test_oauth_state.py -q -n0 --no-cov
- docker compose -f docker-compose.dev.yml --env-file .env exec -T frontend npm run lint -- --file src/app/'(project)'/admin/permissions/page.tsx --file src/components/admin/ProjectPermissionMatrix.tsx --file src/lib/api/permissionApi.ts --file src/app/'(project)'/profile/page.tsx --file src/components/profile/PrivacyExportPanel.tsx --file src/lib/api/profileApi.ts --file src/app/'(auth)'/auth/google/callback/page.tsx --file src/app/slack/callback/page.tsx

Note: DB-backed pytest suites were blocked locally because the configured PostgreSQL test user cannot create databases.


quanwangniuniu (changes requested):
**1. Is the new `project_permission_matrix` endpoint missing authorization checks?**
`project_permission_matrix` in `views.py` is decorated only with `@api_view(['GET'])` — no explicit `permission_classes`, and no check that the requesting user belongs to the project's organization or has admin rights to view its permission matrix.
- *Test:* Call `/api/access_control/projects/<project_id>/permission-matrix/` as an authenticated user from a *different* organization and confirm whether they can see another org's roles/permissions (potential IDOR/cross-tenant leak).
- *Why it matters:* This surfaces role names, permission grants, and org IDs — sensitive structural data — for any project ID, not just ones the caller has access to.

**2. Does the OAuth single-use nonce depend on a shared cache backend that may not exist in production?**
The rewritten `core/services/oauth_state.py` stores nonces in Django's `cache` framework to enforce single-use/expiry, replacing the old stateless `signing.dumps`/`loads` approach used by all seven integrations (Slack, Zoom, Linear, Notion, Google Docs/Calendar/Auth, Facebook).
- *Test:* Confirm what `CACHES["default"]` actually points to in staging/production (Redis vs. per-process LocMemCache). If it's LocMemCache behind multiple app workers/pods, nonce writes and reads could land on different processes, causing valid OAuth callbacks to intermittently fail with `OAUTH_STATE_EXPIRED`.
- *Why it matters:* This single change now gates login/auth for every third-party integration; a cache misconfiguration would cause widespread, hard-to-reproduce auth failures.

**3. Does removing legacy session-based OAuth state break in-flight logins during deploy?**
`test_callback_accepts_legacy_session_state_mode` was renamed to `test_callback_rejects_legacy_session_state_mode` — the fallback that validated state against the Django session cookie was deleted, making signed state the only accepted mechanism.
- *Test:* Start a Google OAuth login just before a deploy that ships this change, and complete the callback just after. Confirm whether users mid-flow get hard failures rather than a graceful fallback.
- *Why it matters:* This is a breaking behavior change for any user with an OAuth flow open at deploy time, with no apparent migration/dual-acceptance window.

**4. Does the GDPR personal-data export over-collect or leak data via broad FK reflection?**
`_query_user_related_rows` iterates over **every** registered Django model (`apps.get_models()`) and builds a `Q` filter for any FK/M2M field pointing to `User`, dumping up to 10,000 rows per model into the export with a generic `except Exception` swallowing errors into the payload.
- *Test:* Audit whether any model with a user FK also contains other users' data in the same row (e.g., a `ChatParticipant` or `ProjectMember` row could expose collaborator names/IDs alongside the requester's), and confirm the 10,000-row cap doesn't silently truncate large accounts without any warning to the user.
- *Why it matters:* GDPR exports should be scoped precisely to the subject's own data — reflection-based blanket collection risks exporting third-party personal data or silently incomplete exports.

**5. Is the export download endpoint's dual "unauthenticated + token" model safe against replay/enumeration?**
`PersonalDataExportDownloadView` uses `AllowAny` and empty `authentication_classes`, relying solely on `verify_download_token`, which checks a signed token but the token itself never expires independently of `export_request.expires_at` (24h) and is not single-use (no nonce/consume step, unlike the new OAuth state design).
- *Test:* Confirm whether the same download token can be reused arbitrarily many times within the 24-hour window (e.g., if leaked via referrer headers, logs, or browser history) and whether there's any rate limiting on this public endpoint.
- *Why it matters:* Given the same PR just built single-use, cache-backed tokens for OAuth state, the export download token uses a weaker, replayable pattern for what is arguably more sensitive data (a full personal data ZIP).

qiwangai:
> **1. Is the new `project_permission_matrix` endpoint missing authorization checks?** `project_permission_matrix` in `views.py` is decorated only with `@api_view(['GET'])` — no explicit `permission_classes`, and no check that the requesting user belongs to the project's organization or has admin rights to view its permission matrix.
> 
> * _Test:_ Call `/api/access_control/projects/<project_id>/permission-matrix/` as an authenticated user from a _different_ organization and confirm whether they can see another org's roles/permissions (potential IDOR/cross-tenant leak).
> * _Why it matters:_ This surfaces role names, permission grants, and org IDs — sensitive structural data — for any project ID, not just ones the caller has access to.
> 
> **2. Does the OAuth single-use nonce depend on a shared cache backend that may not exist in production?** The rewritten `core/services/oauth_state.py` stores nonces in Django's `cache` framework to enforce single-use/expiry, replacing the old stateless `signing.dumps`/`loads` approach used by all seven integrations (Slack, Zoom, Linear, Notion, Google Docs/Calendar/Auth, Facebook).
> 
> * _Test:_ Confirm what `CACHES["default"]` actually points to in staging/production (Redis vs. per-process LocMemCache). If it's LocMemCache behind multiple app workers/pods, nonce writes and reads could land on different processes, causing valid OAuth callbacks to intermittently fail with `OAUTH_STATE_EXPIRED`.
> * _Why it matters:_ This single change now gates login/auth for every third-party integration; a cache misconfiguration would cause widespread, hard-to-reproduce auth failures.
> 
> **3. Does removing legacy session-based OAuth state break in-flight logins during deploy?** `test_callback_accepts_legacy_session_state_mode` was renamed to `test_callback_rejects_legacy_session_state_mode` — the fallback that validated state against the Django session cookie was deleted, making signed state the only accepted mechanism.
> 
> * _Test:_ Start a Google OAuth login just before a deploy that ships this change, and complete the callback just after. Confirm whether users mid-flow get hard failures rather than a graceful fallback.
> * _Why it matters:_ This is a breaking behavior change for any user with an OAuth flow open at deploy time, with no apparent migration/dual-acceptance window.
> 
> **4. Does the GDPR personal-data export over-collect or leak data via broad FK reflection?** `_query_user_related_rows` iterates over **every** registered Django model (`apps.get_models()`) and builds a `Q` filter for any FK/M2M field pointing to `User`, dumping up to 10,000 rows per model into the export with a generic `except Exception` swallowing errors into the payload.
> 
> * _Test:_ Audit whether any model with a user FK also contains other users' data in the same row (e.g., a `ChatParticipant` or `ProjectMember` row could expose collaborator names/IDs alongside the requester's), and confirm the 10,000-row cap doesn't silently truncate large accounts without any warning to the user.
> * _Why it matters:_ GDPR exports should be scoped precisely to the subject's own data — reflection-based blanket collection risks exporting third-party personal data or silently incomplete exports.
> 
> **5. Is the export download endpoint's dual "unauthenticated + token" model safe against replay/enumeration?** `PersonalDataExportDownloadView` uses `AllowAny` and empty `authentication_classes`, relying solely on `verify_download_token`, which checks a signed token but the token itself never expires independently of `export_request.expires_at` (24h) and is not single-use (no nonce/consume step, unlike the new OAuth state design).
> 
> * _Test:_ Confirm whether the same download token can be reused arbitrarily many times within the 24-hour window (e.g., if leaked via referrer headers, logs, or browser history) and whether there's any rate limiting on this public endpoint.
> * _Why it matters:_ Given the same PR just built single-use, cache-backed tokens for OAuth state, the export download token uses a weaker, replayable pattern for what is arguably more sensitive data (a full personal data ZIP).

Thanks, I addressed these in the latest update:
- Added auth + same-org admin checks for the permission matrix endpoint, with a cross-org 403 test.
- Confirmed OAuth state uses Redis-backed Django cache by default; staging/prod should keep Redis for shared nonce storage.
- Added a one-time legacy Google OAuth session-state grace path for in-flight deploy callbacks.
- Tightened GDPR export scope with an explicit model allowlist, subject-user redaction, and truncation metadata.
- Made export download signed URLs single-use with a consumed nonce and added replay coverage.

quanwangniuniu (approved):
Good job!

PR #767, opened 2026-07-23
MED-293: Spreadsheet real-time multi-user editing via Channels

zizhenliu0427:
## Summary

Completes MED-293 — real-time multi-user spreadsheet editing over the existing Django Channels infrastructure. Users working on the same sheet can see presence avatars, remote cursors/selections, committed cell edits, formula-dependent recalculations, and structural refreshes without manually reloading the page.

- **Sheet WebSocket rooms** — each sheet has an isolated, tenant-scoped Channels group at `ws/spreadsheets/sheets/<sheet_id>/`.
- **Authentication and access** — the client obtains a 30-second one-time WebSocket ticket from an authorised REST endpoint. The ticket is bound to the user, sheet, tenant, and per-tab `client_id`, then atomically consumed with Redis `GETDEL`. Legacy spreadsheet JWT query-string connections are rejected.
- **Permission lifecycle** — the consumer runs a 10-second watchdog, rechecks active project membership every 30 seconds, and limits each connection to a configurable maximum of one hour.
- **Presence** — new connections receive a presence snapshot and peers receive join/leave events. Presence is tracked per browser-tab `client_id`, including duplicate-client disconnect protection.
- **Atomic presence cleanup** — Redis Hash and Sorted Set state is updated through Lua scripts across workers. A 10-second heartbeat and 30-second stale-entry window prune abandoned memberships while tolerating short network interruptions.
- **Browser recovery** — `visibilitychange`, `focus`, `online`, and `pageshow` rebuild suspended or interrupted connections. Reconnect attempts use exponential back-off with jitter.
- **Remote cursors** — selection changes are throttled and broadcast to peers. A server-side token bucket limits cursor traffic, and the grid renders a visible remote-cell outline without injecting username text into cell content.
- **Per-cell LWW** — the latest committed REST write wins for a cell. Cell and structural mutations share a per-sheet database row lock, so lock and commit order provide deterministic last-write-wins behaviour across processes.
- **Optimistic editing** — the originating grid updates immediately through the existing edit path. Remote committed values are applied without overwriting a newer pending local edit.
- **Formula consistency** — recalculated dependent cells are included in the same `cells_updated` event as the source edit. Recalculation remains inside the same transaction, with recursive dependency discovery, cached reference reads, and bulk persistence.
- **Echo suppression** — REST writes carry the sheet `client_id`; the server does not send committed cell or refresh events back to the originating connection. The frontend also ignores its own origin defensively.
- **Structure synchronisation** — row/column insert, delete, reorder, sort, resize, revert, and import finalisation advance `Sheet.revision` and broadcast `sheet_refresh_required` with the committed revision.
- **Revision conflict protection** — REST mutations carry `base_revision`. The backend validates it after acquiring the sheet lock and rejects stale-coordinate writes with HTTP 409 and `SHEET_REVISION_CONFLICT`.
- **Canonical refresh hardening** — revision gaps and structural events pause saves, clear coordinate caches, and reload canonical state. Stale queued edits and active editors are discarded with an explicit message, while pointer, keyboard, copy, and paste input remain blocked during the reload.
- **Sheet-tab synchronisation** — sheet create, rename, and delete fan out to every active room in the spreadsheet, including the deleted sheet's room.
- **Concurrent-save safety** — newer optimistic edits remain queued when an earlier REST save completes, while same-sheet writes are serialised to preserve deterministic LWW order.
- **Client isolation and errors** — ticket identity, positive sheet IDs, `client_id` length, tenant schema, and event sheet ownership are validated. REST spreadsheet lookups use project-access-aware helpers, and `AllowedHostsOriginValidator` protects WebSocket upgrades.

## Architecture / Protocol

```text
Local edit
  → existing REST mutation with client_id + base_revision
  → transaction + per-sheet select_for_update lock
  → cell write + formula recalculation
  → transaction.on_commit
  → Channels group_send(cells_updated)
  → peers apply the committed revision
```

WebSocket ticket:

```text
POST /api/spreadsheet/sheets/<sheet_id>/ws-ticket/
```

WebSocket URL:

```text
ws/spreadsheets/sheets/<sheet_id>/?ticket=<one-time-ticket>&client_id=<tab-client-id>
```

Main event types:

- `presence_snapshot`
- `presence_join`
- `presence_leave`
- `cursor_update` / `cursor_updated`
- `cells_updated`
- `sheet_refresh_required`

The implementation deliberately keeps writes on REST instead of adding a second WebSocket write path. This preserves existing authorisation, validation, transaction handling, formula recalculation, and import behaviour in one place.

## Files Changed

(46 files; key areas only)

**Backend — Channels, authentication, presence, consistency, and tests**

- `backend/asset/middleware.py` — one-time spreadsheet ticket authentication
- `backend/backend/asgi.py` — spreadsheet routing and origin validation
- `backend/spreadsheet/consumers.py` — room lifecycle, permission watchdog, presence, cursor, and committed-event handlers
- `backend/spreadsheet/presence.py` — atomic Redis/Lua presence storage and pruning
- `backend/spreadsheet/ws_tickets.py` — ticket minting, binding, expiry, and atomic consumption
- `backend/spreadsheet/tenant.py` — tenant validation and isolated room naming
- `backend/spreadsheet/services.py` — per-sheet locking, LWW broadcasts, revision checks, and recalculation optimisation
- `backend/spreadsheet/views.py` — authorised REST lookups, ticket endpoint, revision metadata, and structural events
- `backend/spreadsheet/models.py` / `backend/spreadsheet/migrations/0011_sheet_revision.py` — persisted `Sheet.revision`
- `backend/spreadsheet/tests/` — conflict, ticket, presence, consumer, and formula performance coverage

**Frontend — socket state, presence, revisions, and grid reconciliation**

- `frontend/src/hooks/useSheetSocket.ts` — ticket connection lifecycle, heartbeat, recovery, back-off, and revision-gap handling
- `frontend/src/lib/sheetSocketStore.ts` — per-sheet socket, presence, and cursor state
- `frontend/src/lib/sheetRevisionStore.ts` — revision tracking, `base_revision`, and 409 conflict recognition
- `frontend/src/components/spreadsheets/presence/` — presence avatars and grid bridge
- `frontend/src/components/spreadsheets/SpreadsheetGrid.tsx` — remote cells, cursor overlay, pending-edit protection, and canonical refresh guards
- `frontend/src/lib/api/spreadsheetApi.ts` — collaboration metadata, ticket requests, and revision responses

**E2E**

- `frontend/e2e/spreadsheets/spreadsheet-presence.spec.ts` — two-tab presence avatar and remote cursor
- `frontend/e2e/spreadsheets/spreadsheet-collab-edit.spec.ts` — peer edit latency, revision, and structure refresh

**Performance validation / QA artefact**

- `k6/scripts/scenarios/spreadsheet-realtime-test.js` — authenticated WebSocket and distributed-edit load scenarios
- `PERFORMANCE-REPORT-MED-293.md` — local QA performance report retained with the submission artefacts and not committed in this PR

## Expected Result

- Two active sessions on the same sheet see each other's committed edits without reloading.
- Peer DOM updates are observed within the ticket's 300 ms target.
- Users see who else is present and where another participant is selecting.
- Concurrent edits to the same cell resolve deterministically: the later committed write wins.
- Formula results and structural changes converge across open sessions.
- Stale writes based on an older sheet structure are rejected before reaching the database.
- A reconnect refreshes canonical state after any missed broadcast window.
- Users removed from the project lose an established socket within the authorisation recheck window.
- Existing single-user spreadsheet editing remains on the established REST path.

## Test

- [x] Backend spreadsheet regression suite — **399 passed**
- [x] Frontend full Jest suite — **893 passed / 126 suites**
- [x] Frontend TypeScript — **0 spreadsheet-related errors**; 4 unrelated existing E2E errors remain
- [x] Current Playwright Chromium collaboration run — **6 passed**
- [x] Targeted canonical structure-refresh Playwright scenario — **passed**
- [x] Measured Playwright peer-edit propagation — approximately **142–148 ms**
- [x] Local k6 distributed two-editor test — **0% REST/WebSocket failures**, approximately **155 ms propagation p95**
- [x] Changed-file lint — **passed with 0 errors**; existing warnings only
- [x] `git diff --check` — **passed**

### Local performance note

The current GitHub Actions workflows do not execute k6. These results were produced manually against the local Docker development stack using `k6/scripts/scenarios/spreadsheet-realtime-test.js`. The two-user ticket acceptance target passed. Local results are not a staging or production capacity certification.

### Local full-backend note

The current spreadsheet backend suite passes in a clean single-worker run: **399 passed**. Frontend Jest and the collaboration Playwright scenarios also pass. The remaining four TypeScript errors are in unrelated existing E2E specifications outside the spreadsheet implementation.

### Migration note

This implementation adds `backend/spreadsheet/migrations/0011_sheet_revision.py` despite the original ticket stating “Migrations: no”. The persisted revision is required to reject stale-coordinate writes atomically while holding the database row lock.

Developers and deployed environments updating this branch must run:

```bash
python manage.py migrate
```

## Out of Scope / Follow-up

- Offline editing and offline event replay
- OT/Yjs/CRDT document synchronisation
- Persisting cursor/presence history
- Automatically rebasing unsaved edits after structural changes
- Asynchronous formula recalculation
- A database-backed monotonic cell/event version for ordering across application nodes with clock skew

Ticket: [MED-293](https://linear.app/mediajira/issue/MED-293/spreadsheet-real-time-multi-user-editing-via-channels)


quanwangniuniu (changes requested):
I reviewed the PR summary and files changed (notably backend/spreadsheet/consumers.py, services.py, routing.py, asgi.py, and the REST batch-update integration). Below are six core review questions — each includes the potential issue I see and a requested solution or clarification.

Authentication & token handling
Question: consumers.py accepts JWT via URL query (ws/.../?token=<jwt>&client_id=...). How do you mitigate token leakage (referer logs, intermediary proxies) and ensure token expiry/revocation is handled on reconnects?
Potential issue: tokens in URLs are more likely to leak; expired/revoked tokens may let disconnected clients rejoin until token expires; replay attacks or stale permissions could allow unauthorized room joins.
Ask for solution: consider moving auth into the WebSocket subprotocol/headers during upgrade or use short-lived connection tokens minted server-side (one-time or tied to client_id). Document token lifecycle and show server-side checks for expiration/revocation on every connect.

Presence, heartbeats, and duplicate-client handling
Question: presence uses a 10s heartbeat + 30s stale window and per-tab client_id with duplicate-client disconnect protection. How do you handle network jitter, browser suspend/resume, and multi-worker race conditions so presence doesn't flicker or leave ghosts?
Potential issue: false-positive stale pruning during short network hiccups; race between Channels worker processes updating/removing presence leading to transient ghost avatars or premature leave events.
Ask for solution: show exact Redis/key TTL design (atomic increments or SETEX), and add a diagram or tests for these races. Consider using redis TTLs + touching on heartbeat; document behavior under suspend/resume and across ASGI worker restarts.

Echo suppression, LWW, and concurrent-save ordering
Question: echo suppression depends on REST writes carrying client_id and server not broadcasting back to that connection. How is the authoritative ordering (LWW) guaranteed across concurrent REST writes from multiple clients and across server processes?
Potential issue: races where REST saves are received out-of-order or a client's optimistic local edits get overridden unexpectedly; edge-case where originating connection disconnects before suppression and still receives committed event; serialization logic might be incomplete.
Ask for solution: describe the serialization mechanism (e.g., DB timestamp ordering, per-sheet write queue, or transactional sequence numbers). Add unit/integration tests that simulate concurrent saves from different client_ids and verify deterministic LWW behavior and no lost optimistic updates.

Formula recalculation atomicity, latency, and scalability
Question: cells_updated includes recalculated dependent cells in the same event. Is recalculation done inside the same DB transaction synchronous with REST batch-update, and what are the latency/scalability implications for large sheets?
Potential issue: long-running recalculation inside a transaction can block or time out under heavy load; partial broadcasts if the recalculation span is handled async and events fire prematurely; inconsistent state if recalculation is offloaded but events aren't reconciled.
Ask for solution: clarify whether recalculation is fully atomic in the REST transaction and emitted on transaction.on_commit; if heavy recalculation is possible, provide mitigation (batching, worker offload pattern with revision numbers, or incremental diffs) and add performance tests showing end-to-end latency bounds.

Structure synchronization, coordinate caching, and refresh UX
Question: structural ops broadcast sheet_refresh_required forcing peers to reload canonical state. How do clients reconcile ongoing optimistic edits, and how is coordinate cache invalidation handled to avoid mapping errors (stale cell coords) after row/column changes?
Potential issue: refresh can interrupt in-flight edits and cause data loss or visual jumps; stale local caches may map cell IDs incorrectly leading to incorrect edits being applied post-refresh.
Ask for solution: introduce a sheet_revision or op-version in the refresh event so clients can deterministically reconcile optimistic edits, and define a clear client flow (pause local saves, rebase/reapply optimistic queue, then resume). Add tests for reapply/rebase correctness across structural changes.

Security, room isolation, and broadcast safeguards
Question: services.py constructs room names and broadcasts events. What server-side validation ensures cross-sheet or cross-project events cannot be injected or received by the wrong room (e.g., malformed sheet_id, path traversal, incorrect group_name construction)?
Potential issue: incorrect room naming or lack of validation could allow events to leak across rooms or allow a malformed request to broadcast to an unintended group; lack of per-event authorization checks risks broadcasting sensitive changes to unauthorized clients.
Ask for solution: require canonical validation at broadcaster entrypoints (verify sheet_id exists and the caller has project membership) and include defensive assertions in services.py and consumers.py. Provide tests that attempt to broadcast to wrong rooms and verify broadcasts are rejected.

zizhenliu0427:
## Review questions and implemented solutions

### 1. Authentication and token lifecycle — fully followed

**Implementation**

- The main changes are in `backend/spreadsheet/ws_tickets.py`, `backend/asset/middleware.py` and the connection lifecycle in `backend/spreadsheet/consumers.py`.
- Added a REST endpoint that issues a 30-second, cryptographically random WebSocket ticket.
- Each ticket is bound to the user, sheet, `client_id`, tenant schema and token/connection expiry.
- Tickets are consumed atomically with Redis `GETDEL`, so they cannot be replayed.
- Spreadsheet WebSocket routes reject the legacy JWT query-string authentication path.
- The connection runs a 10-second authorisation watchdog and rechecks active project membership against the database every 30 seconds.
- A connection is limited to a maximum of one hour, even when the originating authentication session would otherwise last longer.
- The ticket endpoint is rate-limited to 60 requests per minute.
- Reconnection always returns through the authorised ticket endpoint, so expiry and current access are checked on every new connection.

**Trade-off**

An established socket does not query the JWT blacklist on every message. That would add database or cache work to the hottest path. Permission removal is instead enforced within the 30-second membership-check window, and reconnection always requires a new authorised ticket.

### 2. Presence, network jitter and multiple workers — fully followed

**Implementation**

- `backend/spreadsheet/presence.py` now owns the atomic Redis operations, while `frontend/src/hooks/useSheetSocket.ts` handles heartbeat and browser recovery.
- Replaced cache `get → modify → set` presence updates with Redis Hash and Sorted Set storage.
- Registration, heartbeat/touch, removal, duplicate-client replacement and stale pruning use Lua scripts so each operation is atomic across ASGI workers.
- The client sends a heartbeat every 10 seconds; entries become stale after 30 seconds.
- Stale pruning publishes an authoritative presence snapshot, removing ghost avatars from every connected client.
- Presence identity is per user and per browser tab through `client_id`.
- On `visibilitychange`, `focus`, `online` and `pageshow`, a suspended or restored page rebuilds the connection.
- Reconnection uses exponential back-off with jitter to prevent many tabs reconnecting simultaneously.
- Concurrency tests cover simultaneous registration, stale pruning and duplicate-client removal, satisfying the review's request for race documentation or tests.

**Trade-off**

A disconnected user may remain visible for up to roughly 30 seconds. This grace period is intentional: removing users immediately after one delayed heartbeat would cause visible flicker during brief network jitter or browser scheduling pauses.

### 3. Echo suppression and deterministic LWW ordering — fully followed

**Implementation**

- The core ordering and broadcasts are implemented in `backend/spreadsheet/services.py`; optimistic-event handling is in `SpreadsheetGrid.tsx`.
- Every tab has its own `client_id`, included in REST mutations and resulting WebSocket events.
- The consumer suppresses committed cell and refresh events for the originating connection; the frontend also ignores its own origin as a defensive second layer.
- Cell mutations and coordinate-changing structural mutations share a `select_for_update()` lock on the `Sheet` row inside `transaction.atomic`.
- The database lock serialises concurrent writers across processes. The later lock/commit wins, giving deterministic last-write-wins behaviour.
- Broadcasts run only through `transaction.on_commit`, so clients never receive uncommitted or rolled-back values.
- Remote cell events do not overwrite a cell that still has a pending local optimistic operation.
- Real concurrent database tests verify that the second writer waits for the first lock, the later committed value wins and stale structural writes are rejected.

**Trade-off**

The lock is per sheet rather than per cell. This is simpler and also prevents cell writes from interleaving with row or column changes, but two unrelated cell edits on the same sheet are briefly serialised. The current latency results show this is acceptable for the ticket's expected load.

### 4. Formula recalculation — fully followed through batching

**Implementation**

- The main changes are in `backend/spreadsheet/services.py` and `backend/spreadsheet/formula_engine.py`.
- Dependent formula recalculation runs synchronously in the same transaction as the source-cell edit.
- The source edit, recalculated dependants and revision are broadcast only after the transaction commits.
- Dependency traversal uses a recursive CTE on PostgreSQL, with a frontier-based breadth-first fallback.
- Formula evaluation uses a per-recalculation cell snapshot/cache instead of repeatedly querying references.
- Recalculated cells are persisted with `bulk_update` rather than one `save()` call per cell.
- A 100-cell dependency-chain performance test enforces a query budget below 300 and an execution budget below five seconds in the test environment.

**Trade-off**

Recalculation remains synchronous to preserve one atomic visible state. Very large formula graphs can hold the sheet lock for longer. Moving recalculation to a worker would require a separate two-phase revision and reconciliation protocol, so asynchronous offload remains a future scaling option rather than part of this ticket.

The review offered batching, worker offload or incremental diffs as alternative mitigations. This implementation follows the batching/optimisation option; worker offload was not required in addition.

### 5. Structural changes and coordinate safety — partially followed

**Implementation**

- The revision protocol is implemented across `Sheet.revision`, the REST mutation services, `sheetRevisionStore.ts` and `SpreadsheetGrid.tsx`.
- Added `Sheet.revision`, advanced while the common sheet lock is held whenever a structural operation changes coordinate meaning.
- REST mutations carry `base_revision`; responses and WebSocket events carry the current `revision`.
- A mutation based on stale structure is rejected with HTTP 409 and `SHEET_REVISION_CONFLICT`.
- The frontend recognises either the business error code or an HTTP 409 containing a valid `current_revision`.
- A revision gap or structural refresh pauses saves, waits for in-flight work, clears coordinate-dependent caches and reloads canonical sheet state.
- Queued optimistic edits and any active editor tied to the old coordinates are cancelled with an explicit message to the user.
- Pointer, keyboard, copy and paste input are disabled while canonical data is loading.
- The edit commit handler has its own in-flight refresh guard, covering Enter and blur events from an editor opened before the refresh.

**Trade-off**

The client safely discards stale-coordinate edits rather than attempting automatic rebase. A correct rebase would need to reproduce the backend's row/column and formula-reference rewrite rules in the frontend. The current behaviour favours correctness and a clear warning over silently applying an edit to the wrong cell.

The literal request to “rebase/reapply optimistic queue” is therefore not implemented. Tests cover revision conflicts, cache invalidation, interaction blocking and safe discard rather than rebase correctness.

### 6. Security, tenant isolation and broadcaster safeguards — partially followed

**Implementation**

- The critical changes are the authorised REST lookup helpers, ticket-bound consumer access and tenant-aware room naming in `backend/spreadsheet/tenant.py`.
- REST spreadsheet lookups now consistently use project-access-aware helpers.
- WebSocket connection authorisation uses the same active project membership policy.
- Ticket, URL sheet ID and `client_id` must match.
- Sheet IDs must be positive integers and `client_id` length is bounded.
- Non-public channel group names include a SHA-256-derived tenant-schema component, preventing identically numbered sheets in different tenants from sharing a room.
- Tenant schema names are validated before they are used for ORM access or room construction.
- Broadcasters reject malformed sheet IDs and cell payloads whose cells do not belong to the target sheet.
- Cursor messages are protected by a token-bucket rate limiter.
- `AllowedHostsOriginValidator` protects the WebSocket upgrade against untrusted browser origins.

**Trade-off**

The internal broadcast helpers do not repeat project-membership queries. All externally reachable mutations already pass through the authorised REST boundary, and repeating those queries would add two or three database queries to every edit batch. A broadcast to a valid but empty/non-existent room is harmless; malformed input and cross-sheet cell payloads still fail loudly.

The literal request to verify sheet existence and caller membership again at each broadcaster entry point is therefore not implemented. Equivalent protection is applied at the external REST mutation and WebSocket connection boundaries, while broadcaster tests cover malformed targets, cross-sheet payloads and tenant isolation.

## Latest recorded validation

- Backend spreadsheet tests: **399 passed**.
- Frontend Jest: **126 suites, 893 tests passed**.
- Spreadsheet-related TypeScript errors: **0**. Four unrelated existing E2E type errors remain elsewhere in the repository.
- Targeted Playwright canonical structure-refresh scenario: **passed**.
- Earlier full collaboration Playwright run: **6 passed**.
- Measured peer-edit propagation: approximately **142–148 ms** in Playwright.
- k6 distributed scenario: **0% write/WebSocket failures**, propagation p95 approximately **155 ms**.

## Known limitations and follow-up items

- `Sheet.revision` requires `backend/spreadsheet/migrations/0011_sheet_revision.py`, despite the ticket originally stating “Migrations: no”. The field is required by the implemented stale-coordinate protection; deployments and developers updating the branch must run the normal **Django migration** command.
- Cell-event ordering still exposes Python-generated `updated_at` to the frontend. The database row lock preserves the authoritative final state, but significant clock skew between application nodes could cause the UI to reject a later event with an earlier timestamp. A database-backed monotonic cell/event version would be the stronger long-term solution.
- Automatic rebase of unsaved edits after structural changes and asynchronous formula recalculation are intentionally outside the current implementation.

quanwangniuniu (approved):
Approved!

PR #772, opened 2026-08-05
feat(access-control): add admin override audit log (MED-338)

Joestar577:
## Summary

Implements MED-338 by adding an audit trail for superuser and organization-admin permission bypasses. `AuthorizationMiddleware` already let superusers and org admins (role level 2) skip module-level RBAC checks, but nothing recorded when that happened — there was no way to see who used elevated access, on what module/action, or why. This PR adds an `AdminOverrideAudit` model that logs each bypass (user, organization, override type, module, action, HTTP method/path, optional reason header, IP address), a read-only API to list those entries, and an admin-only frontend page to browse and filter the log.

The middleware change reorders the existing bypass logic rather than adding new gates: module/action parsing (previously done only after the bypass checks) now happens first, so the middleware can tell whether a real permission gate would have applied before deciding to log — this avoids noise from paths like `/api/auth/login/` that never reach a module check. Logging is best-effort and wrapped in `try/except` so a logging failure can never block a request the bypass already allowed.

## Backend

- `access_control/models.py`: add `AdminOverrideAudit` (`TimeStampedModel`) with `user`, `organization` (nullable, `SET_NULL`), `override_type` (`SUPERUSER`/`ORG_ADMIN`), `module`, `action`, `method`, `path`, `reason`, `ip_address`; indexed on `-created_at` and `(user, -created_at)`.
- `access_control/middleware/authorization.py`: hoist module/action path parsing above the superuser/org-admin bypass checks in `AuthorizationMiddleware`; add `_log_override()`, called from both bypass branches only when `has_permission_gate` is true.
- `access_control/views.py`: add `admin_override_audit_list` (`GET /api/access-control/admin-override-audits/`), restricted to superusers and org admins (`is_org_admin`), with `user_id`, `override_type`, `module`, and `from`/`to` (ISO 8601, filtering `created_at`) query params, paginated via `AdminOverrideAuditPagination` (50/page, max 200).
- `access_control/serializers.py`: add read-only `AdminOverrideAuditSerializer`.
- `access_control/urls.py`: register `admin-override-audits/` route.
- `core/tenant_config.py`: register `AdminOverrideAudit` as a tenant model (depends only on `User` + `Organization`, same tier as `Role`).
- `access_control/migrations/0003_adminoverrideaudit.py`: creates the new table and its two indexes.

## Frontend

- New `/admin/override-audits` page (added under both `(project)/admin/override-audits/page.tsx` and the `[orgSlug]/[projectSlug]/admin/override-audits/page.tsx` re-export wrapper, matching this app's dual URL-scheme routing).
- `components/audit/OverrideAuditTable.tsx`, `OverrideAuditEntryRow.tsx`, `OverrideAuditFilterBar.tsx`: table/row/filter UI for browsing the audit log.
- `lib/api/adminOverrideAuditApi.ts` + `types/adminOverrideAudit.ts`: typed API client and response types for the new endpoint.
- `DashboardSidebar.tsx` and `Sidebar.tsx`: add an "Override Audit Log" entry (History icon) to the admin nav group, linking to `/admin/override-audits`.
- `nginx/nginx.conf`: add `override-audits` to the Next.js `/admin/*` route allowlist regex so the new page isn't shadowed by the Django `/admin/` prefix.

## Tests

    - `docker compose -f docker-compose.dev.yml --env-file .env exec backend pytest access_control/tests/test_admin_override_audit.py access_control/tests/test_admin_override_audit_view.py --no-cov`
    - Result: 18 passed
    - (`--no-cov` used to avoid unrelated global coverage-threshold failures on a targeted run)

- Frontend targeted unit tests:
    - `docker compose -f docker-compose.dev.yml --env-file .env exec frontend npx jest OverrideAuditEntryRow OverrideAuditFilterBar OverrideAuditTable adminOverrideAuditApi`
    - Result: 33 passed (4 suites)

- Frontend TypeScript check:
    - `docker compose -f docker-compose.dev.yml --env-file .env exec frontend npx tsc --noEmit`
    - Result: clean for all files touched by this PR. Remaining errors are pre-existing and confined to `e2e/` specs unrelated to this ticket (`oauth.spec.ts`, `sidebar-navigation.spec.ts`, `preference-refetch-on-focus.spec.ts`, `task-parent-cycle.spec.ts`).

- Diff hygiene:
    - `git diff --check`
    - Result: clean

## QA

- Demo video and QA checklist submitted in Discord `#qa-feedback`; approved by Ray.

## Notes

- Migrations: yes. `backend/access_control/migrations/0003_adminoverrideaudit.py` (new table, no data migration).
- Audit logging is best-effort: `_log_override()` swallows exceptions so a logging failure never blocks the request the bypass already allowed. A failed write is currently silent (no error-tracking hook) — acceptable for this PR's scope but worth flagging for future hardening.
- The override log endpoint is intentionally read-only and gated to superusers/org admins only, since the log itself is sensitive access-control data.


quanwangniuniu (commented):
---

## 1. **Race Condition: Organization Context Inconsistency**

**Issue:** The middleware logs `organization_id` using `getattr(user, 'current_organization_id', None) or getattr(user, 'organization_id', None)`, but this may not reflect the actual tenant schema being queried. In a multi-tenant system, a user making a request in one organization could have their `current_organization_id` pointing elsewhere.

**Impact:** Audit logs may show the wrong organization for an override event.

**Solution:**
```python
# In AuthorizationMiddleware._log_override()
# Extract org from request context headers instead of relying on user state
org_id = request.META.get('HTTP_X_ORGANIZATION_ID')
if not org_id and hasattr(request, 'tenant'):
    org_id = request.tenant.id
AdminOverrideAudit.objects.create(
    organization_id=org_id,  # Use request context, not user.current_organization_id
    ...
)
```

---

## 2. **Path Parsing Logic Too Fragile**

**Issue:** The module/action extraction uses naive string manipulation:
```python
raw = parts[1]
candidate_module = raw.rstrip('s').upper()  # 'assets' → 'ASSET'
```
This breaks for:
- Plurals that don't end in 's' (e.g., `campaigns` → `campaign` ✓ but what about `data`?)
- Actual modules containing 's' (e.g., if a future module is "ASSETS" singular, `rstrip('s')` breaks it)
- Non-RESTful paths like `/api/bulk-operations/approve`

**Impact:** Audit logging may silently miss overrides on certain paths.

**Solution:**
```python
# Define a mapping instead of string manipulation
MODULE_PATH_MAP = {
    'assets': 'ASSET',
    'campaigns': 'CAMPAIGN',
    'budgets': 'BUDGET',
    'reporting': 'REPORTING',
}
candidate_module = MODULE_PATH_MAP.get(raw.lower())
# Fail safely: if not in map, don't create audit entry
```

---

## 3. **No Rate-Limiting or DDoS Protection on Audit Logging**

**Issue:** A malicious superuser could spam override requests to flood the audit log. The code has `try/except` for logging failures, but no throttling or batch-write protection.

**Impact:** 
- Audit table grows unbounded
- Query performance degrades as filters scan millions of rows
- Storage costs balloon
- The audit purpose (accountability) is defeated by noise

**Solution:**
```python
# Add retention policy and rate limiting
class AdminOverrideAudit(TimeStampedModel):
    # ... existing fields ...
    
    class Meta:
        # Keep only 90 days of audit data
        indexes = [
            models.Index(fields=['-created_at']),
            models.Index(fields=['user', '-created_at']),
        ]

# Add a periodic task (Celery/APScheduler)
def cleanup_old_audits():
    cutoff = timezone.now() - timedelta(days=90)
    AdminOverrideAudit.objects.filter(created_at__lt=cutoff).delete()
```

---

## 4. **Missing Tests for Edge Cases and Failure Modes**

**Issue:** Current tests cover the happy path but miss:
- What happens if `AdminOverrideAudit.objects.create()` fails due to database issues? The test uses `try/except` to swallow errors silently — no alert mechanism.
- No test for concurrent requests (race condition on org detection)
- No test for malformed paths or deeply nested routes
- No test verifying that normal users cannot access the audit log (only happy-path test exists)

**Impact:** Silent failures could mask audit log gaps without alerting admins.

**Solution:**
```python
# In tests/test_admin_override_audit.py
def test_logging_failure_does_not_block_request(self):
    """Ensure best-effort logging never blocks a valid bypass."""
    with patch.object(AdminOverrideAudit.objects, 'create', side_effect=Exception('DB error')):
        req = self.factory.post('/api/campaigns/create/')
        req.user = self.superuser
        resp = self.middleware.process_view(req, dummy_view, (), {})
        # Request should still succeed even if logging fails
        self.assertIsNone(resp)

def test_deeply_nested_paths_ignored(self):
    """Paths that don't match expected structure should not log."""
    req = self.factory.post('/api/v2/campaigns/nested/deep/create/')
    req.user = self.superuser
    self.middleware.process_view(req, dummy_view, (), {})
    # Only well-formed paths should log
    self.assertEqual(AdminOverrideAuditEntry.objects.count(), 0)
```

---

## 5. **Serializer Missing Null Safety and Audit Trail Gaps**

**Issue:** The `AdminOverrideAuditSerializer` includes:
```python
organization_id = serializers.IntegerField(read_only=True)
```
But `organization` can be `null` (foreign key with `SET_NULL`), so accessing `.organization_id` on a null org will fail silently or return incorrect data. Additionally, there's no `is_deleted` filtering in the list view, so soft-deleted records may leak into the UI.

**Impact:** 
- Frontend may receive `null` or missing `organization_id` unexpectedly
- Deleted audit entries could be visible if someone bypasses the filter

**Solution:**
```python
class AdminOverrideAuditSerializer(serializers.ModelSerializer):
    organization_id = serializers.SerializerMethodField(read_only=True)
    
    def get_organization_id(self, obj):
        return obj.organization.id if obj.organization else None
    
    class Meta:
        model = AdminOverrideAudit
        fields = [
            'id', 'user_id', 'username', 'organization_id', 'override_type',
            'module', 'action', 'method', 'path', 'reason', 'ip_address',
            'created_at',
        ]
        read_only_fields = fields

# In views.py, ensure soft-delete filtering
queryset = AdminOverrideAudit.objects.filter(is_deleted=False).select_related('user', 'organization')
```

---

## Summary

| Issue | Severity | Impact |
|-------|----------|--------|
| Org context mismatch | 🔴 High | Wrong org logged | 
| Path parsing fragile | 🟡 Medium | Missed audits |
| No DDoS protection | 🟡 Medium | Audit table bloat |
| Missing edge-case tests | 🟡 Medium | Silent failures |
| Null safety in serializer | 🟠 Low | Frontend bugs |

All five issues are fixable with targeted code changes. The first two are architectural and should be addressed before merge; the others can be follow-ups.

quanwangniuniu (changes requested):
See the comments above

Joestar577:
Thanks for the detailed review, I agree with these points. Here's my plan:

Issue 1 (org context): Will fix — switching from user.current_organization_id to the actual tenant context from the DB connection.
Issue 2 (path parsing): Will fix — replacing rstrip('s') with an explicit MODULE_PATH_MAP dictionary.
Issues 3, 4, 5: Will create follow-up tickets for rate-limiting/retention, extra edge-case tests, and serializer null safety.

Joestar577:
Addressed issues 1 and 2 in `d1a42ed3b`. Issues 3–5 are intentionally left as follow-ups per scope.

**Issue 1 (org context race):** `_log_override()` no longer trusts `user.current_organization_id`. It now reads the DB connection's current `search_path` (already set correctly for this request by `TenantSchemaMiddleware`) and matches it against each organization's tenant schema (`slug_to_schema_name`) to find the actual org being queried. This guarantees the audit row's `organization_id` always matches the tenant schema it's written into, regardless of what the user's cached `current_organization_id` says. Verified against `test_admin_override_audit.py`, which explicitly sets `search_path` to a test org's schema and asserts `audit.organization_id == self.org.id`.

**Issue 2 (fragile path parsing):** Replaced `raw.rstrip('s').upper()` with an explicit `MODULE_PATH_MAP` dict on `AuthorizationMiddleware`. I traced the actual URL routing rather than guessing at the map contents: `assets` → `ASSET` and `campaigns` → `CAMPAIGN` were the only two modules ever reachable through this top-level `/api/<segment>/` check, even under the old logic — budgets/queues/tickets/invitations are all nested one path segment deeper (`/api/budgets/requests/`, `/api/csm/tickets/`, `/api/core/invitations/`) and were never reachable here before either. The map preserves that exact existing behavior; expanding module coverage would be a separate, out-of-scope change.

Ran the full existing suite: `pytest access_control/ --no-cov` → 96 passed, no regressions.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #775, opened 2026-08-06
MED-243: Isolate LLM step timeout from rest of workflow

Popp180:
### What this does

Create per-step retry policy decorators for executors, to avoid one LLM step timeout aborting the workflow. Allow steps to be skipped and make the workflow continue. No migration needed.

### Rationale

- **Per-step policy vs existing global**: In some cases, the workflow can finish without a step's output, such as detect_columns and generate_criteria, while analyze_data and call_llm are required.
- **Per-step configuration vs default value**: The retry policy now reads max_retries, retry_delay, and on_exhausted from step.config first. If a step sets any of these, that value overrides the decorator's default.
- **Fall back path**: If detect_columns is skipped, the next step, normalize_data, falls back to the original column names and passes them to analyze_data. analyze_data already tolerates a missing column_mapping. If generate_criteria is skipped, analyze_data — the next step — already tolerates a missing success_criteria the same way.

### Key changes

**Backend:**

- `backend/agent/executors.py `— wrap LLM call with timeout-scoped retry policy
- `backend/agent/services.py `— checkpoint progress between steps
- `backend/agent/gemini_client.py` — raise a distinct exception when Gemini's own 429 retries are exhausted, so the per-step policy doesn't retry on top of it
- `backend/agent/test_retry_policy.py` — unit tests covering retry, skip, and per-step config override behavior

**Frontend:**

- `frontend/src/components/agent/chat/AgentChatPage.tsx` — render the real step status (e.g. skipped) from the backend instead of always showing "running"
- `frontend/e2e/agent/skip-step-real-session.spec.ts` — end-to-end test for the skip flow

### Out of scope

Cross-workflow retry queue

### Notes
Although this ticket's scope included a Celery-based checkpoint, the workflow actually runs synchronously inside a single HTTP/SSE connection. Since the acceptance criteria didn't require it, I'm noting this as an observation rather than a defect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
-# MediaJira
+# Marketing Simplified
 
 > **The best media buyer Jira platform in the world**
-> MediaJira is a campaign management platform tailored for media buying teams. It streamlines the process of creating, tracking, and optimizing advertising campaigns, while providing collaboration tools, performance analytics, budget control, and professional API access.
+> Marketing Simplified is a campaign management platform tailored for media buying teams. It streamlines the process of creating, tracking, and optimizing advertising campaigns, while providing collaboration tools, performance analytics, budget control, and professional API access.
 
 ---
 
@@ -165,7 +165,7 @@
 - **agent**: Powers AI-assisted workflow orchestration, decision support, and execution guidance.
 - **automationWorkflow**: Defines and executes automation flows across business modules.
 - **decision**: Tracks decision records, approval states, and related governance flows.
-- **integrations** (`notion_editor`, `slack_integration`, `linear_integration`, `zoom_integration`, `google_calendar_integration`, `google_docs_integration`): Connects MediaJira with external systems for synchronized workflows.
+- **integrations** (`notion_editor`, `slack_integration`, `linear_integration`, `zoom_integration`, `google_calendar_integration`, `google_docs_integration`): Connects Marketing Simplified with external systems for synchronized workflows.
 
 ---
 
@@ -201,8 +201,8 @@ Minimum local DB-related values:
 
 ```env
 DB_HOST=host.docker.internal
-POSTGRES_DB=mediajira_db
-POSTGRES_USER=mediajira_user
+POSTGRES_DB=Marketing Simplified_db
+POSTGRES_USER=Marketing Simplified_user
 POSTGRES_PASSWORD=<your-password>
 POSTGRES_PORT=5432
 ```
@@ -271,14 +271,14 @@ docker compose -f docker-compose.dev.yml --env-file .env up --build -d
 
 Each tagged release (`v*`) publishes CycloneDX SBOM files as GitHub Release assets:
 
-- `mediajira-backend-python-<tag>.cdx.json`
-- `mediajira-frontend-npm-<tag>.cdx.json`
+- `Marketing Simplified-backend-python-<tag>.cdx.json`
+- `Marketing Simplified-frontend-npm-<tag>.cdx.json`
 
 The same files are also available as workflow artifacts on the release tag run.
 
 ## 📊 Monitoring & Observability
 
-MediaJira includes monitoring and observability tools for local development and troubleshooting:
+Marketing Simplified includes monitoring and observability tools for local development and troubleshooting:
 
 ### Metrics Collection
 
--- a/backend/agent/executors.py
+++ b/backend/agent/executors.py
@@ -9,6 +9,8 @@ import os
 from django.core.cache import cache
 import time
 import anthropic
+from .gemini_client import GeminiRetriesExhausted
+
 
 logger = logging.getLogger(__name__)
 
@@ -46,6 +48,13 @@ def retry_policy(max_retries=3,  retry_delay= 5,on_exhausted='fail'):
                         else:
                             return StepResult(success=False, error=str(e), skipped=True)
                     else:
+                        logger.warning(
+                            "LLM call failed: %s. Retry %s/%s in %ss.",
+                            e,
+                            attempt + 1,
+                            max_retries - 1,
+                            retry_delay,
+                        )
                         time.sleep(retry_delay)
         return wrapper
     return decorator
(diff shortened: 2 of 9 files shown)


quanwangniuniu (changes requested):
Scope: PR-wide
``` 
Question: The PR introduces a per-step retry policy and isolates LLM step timeouts from the rest of the workflow. How does the new per-step retry interact with any existing global retry/backoff logic? Specifically:
- Can retries be duplicated (both global and per-step) leading to more attempts than intended?
- Are there safeguards to avoid retry storms or duplicated side effects when a step is retried (idempotency concerns)?
- Could this change alter end-to-end latency or resource usage in production?

Could you add a short explanation of the interactions and, if applicable, a unit/integration test that demonstrates the intended behavior when both per-step and global retry policies are present?
```
Why this matters: Understanding retry interactions prevents unexpected duplicate work, increased costs, or inconsistent system state; tests/documentation here make the change safer to review and deploy.

Scope: PR-wide
``` 
Question: When an LLM step is skipped due to timeout (and the "skipped step" message is created), what is the downstream contract? Concretely:
- Which downstream consumers receive the "skipped" signal/message and how should they behave?
- Is the skipped state surfaced in logs/metrics/alerts so operators can investigate timeouts consistently?
- If a skipped step means partial workflow results, is there a clear fallback or remediation path?

Please document the expected downstream behavior and add metrics/logging points (and a short end-to-end test) showing how a skipped LLM step is recorded and how downstream components react.
```
Why this matters: Clear contracts and observability for skipped steps prevent silent failures and make it possible to monitor, debug, and recover from LLM timeouts in production.

Scope: PR-wide
``` 
Question: Why was moving to per-step retry necessary versus adjusting the existing global timeout/retry configuration? For example:
- Is the goal to allow different timeouts/retries per LLM step because they have different SLAs or cost profiles?
- Were there observed incidents that motivated this split (links or short summary would help)?
- Would a configurable default (with per-step override) be safer than a hard change to the control flow?

Please add a short rationale in the PR description and, if not already present, expose the per-step retry settings via config (with sensible defaults) and a migration/compatibility note for operators.
```
Why this matters: A recorded rationale and configurable defaults help reviewers and operators understand the trade-offs (cost, latency, complexity) and reduce surprise when the change is rolled out.

Popp180:
## How does the new per-step retry policy interact with existing global retry logic?

Two global retry layers already exist, one per LLM provider. The new per-step layer sits on top of both.

### Existing global retry policies

1. Gemini — HTTP 429 (rate limit) retries, in backend/agent/gemini_client.py:32. Up to 4 attempts, with backoff of 2.0s, 4.0s, then 8.0s.
2. Claude (Anthropic SDK) — built-in retry, [<https://platform.claude.com/docs/en/cli-sdks-libraries/sdks/python>]. By default, the SDK retries certain errors twice, with exponential backoff: connection errors, HTTP 408, 409, 429, and any 5xx.

### For steps that call Claude directly

I set the per-step retry count to 1, which means this executor totally run once.so nothing is added on internal retries. This avoids duplicate retries, with no extra latency or resource cost.

Unit test: backend/agent/test_retry_policy.py:169. Confirms the per-step layer makes exactly one call to _call_llm, even when the Anthropic SDK ultimately raises APITimeoutError — which only happens after the SDK's own internal retries are exhausted.

### For steps that call Gemini

I made two changes:

1. Added a custom exception, raised when Gemini's own 429 retries are exhausted.
2. Executors now catch this exception and return a step result directly, without going through the per-step retry policy — avoiding a retry storm and the extra latency/cost that would come with it.

Unit test: backend/agent/test_retry_policy.py:116. Runs the real global backoff loop against a mocked 429 response and confirms the per-step layer does not add retries on top of it: exactly 4 HTTP requests occur (matching Gemini's own retry limit), not 4 × 3 = 12.

## When an LLM step is skipped due to timeout (and the "skipped step" message is created), what is the downstream contract?

### Who are downstream consumers receiving signals?

1. Chat UI — frontend/src/components/agent/chat/StepProgress.tsx renders a distinct color and icon per step status; skipped shows a faded circle. The user also gets a message via a text SSE event: 'The "{step.name}" step was skipped after retries. Continuing with the remaining steps.' (services.py:2362–2364).
2. Database — the step's execution record is saved with status='skipped' and completed_at set to now (services.py:2355–2357).
3. Next workflow steps. This will be elaborated in fallback or remediation path below.

### Is there a logs/metrics/alerts for skipped state?

Yes.

1. Each executor logs a warning at the moment its error occur.
2. Each retry attempt itself is logged, with the current retry count and delay before the next try — backend/agent/executors.py:51, inside the retry_policy decorator.
3. The orchestrator logs once more when a step is formally marked skipped, including the workflow run id, step name, and step type — backend/agent/services.py:2359.

### Is there a clear fallback or remediation path?

Yes.

1. If detect_columns is skipped, the next step, normalize_data, falls back to the original column names and passes them to analyze_data. analyze_data already tolerates a missing column_mapping.
2. If generate_criteria is skipped, analyze_data — the next step — already tolerates a missing success_criteria the same way.

## Why was moving to per-step retry necessary versus adjusting the existing global timeout/retry configuration?

### What's the criteria and motivation for per-step policy?

The criteria isn't SLA or cost — it's whether the workflow can finish without a step's output. analyze_data and call_llm are required: the workflow should stop if either fails. detect_columns and generate_criteria are optional: the workflow can finish without them.

No specific incident drove this — it's a preventive fix.

### Would a configurable default (with per-step override) be safer than a hard change to the control flow?

Yes. I made the following change:

* retry_policy now reads max_retries, retry_delay, and on_exhausted from step.config first. If a step sets any of these, that value overrides the decorator's default. No database migration is needed.
* If a step doesn't set them, the existing hardcoded default applies.

Unit test: backend/agent/test_retry_policy.py:70. Sets AnalyzeDataExecutor's max_retries to 1 via step.config and confirms it overrides the decorator's default of 3.

quanwangniuniu (approved):
Good job

PR #776, opened 2026-08-08
fix(task): add preserve concurrent edits in task field history

Driiizzleeeee:
## Summary

Implemented MED-237 to preserve concurrent Task field edits in `TaskFieldHistory`.

Concurrent writes to the same Task field are now recorded as separate history transitions with the correct author, value chain, and stable ordering. The frontend also reconciles optimistic inline edits with the authoritative Task returned by the server.

## Problem

Before this change, when multiple writes updated the same Task field close together, an earlier transition could be overwritten or dropped from `TaskFieldHistory`.

This meant only the last write could be represented correctly, which could hide intermediate changes and make the recorded before/after chain inaccurate.

## Changes

- Updated `backend/task/models.py`
  - Added append-only Task field transition handling
  - Rebased each transition against the latest recorded history value
  - Added stable timestamps to preserve transition ordering
  - Skips transitions that become no-ops after rebasing
  - Keeps Task save and history handling within the database transaction

- Updated `backend/task/signals.py`
  - Captures Task field changes during `pre_save`
  - Stores multiple pending transition batches without overwriting earlier batches
  - Appends captured transitions after the Task save completes
  - Preserves the correct author for each transition batch

- Added `backend/task/tests/test_field_history.py`
  - Simulates two writers using independently loaded Task instances in the same transaction
  - Verifies both Summary transitions are recorded
  - Verifies the correct before/after chain
  - Verifies correct authorship and timestamp ordering
  - Verifies the final Task value matches the last accepted write

- Updated `frontend/src/lib/taskStore.ts`
  - Added `resolveTaskFromServer`
  - Merges the authoritative server Task into the local Task list
  - Reconciles `currentTask` when the updated Task is currently open

- Updated `frontend/src/components/tasks/ListView.tsx`
  - Keeps the existing optimistic inline-edit behavior
  - Reads the successful PATCH response
  - Reconciles optimistic state with the Task returned by the server

- Updated `frontend/src/components/tasks/detail/TaskFieldHistoryBlock.tsx`
  - Added stable History row identifiers for E2E verification
  - Preserves per-row author, previous value, and new value display

- Added `frontend/e2e/tasks/task-history-concurrent.spec.ts`
  - Opens the same Task in two browser tabs
  - Performs two different inline Summary edits
  - Verifies both PATCH requests succeed
  - Verifies both History transitions are present
  - Verifies the transition chain and final server value
  - Reloads both tabs and confirms server truth
  - Verifies the History UI displays the expected author and values

## Testing

- Focused two-writer backend test passed locally  
  `docker compose -f docker-compose.dev.yml exec backend python manage.py test task.tests.test_field_history.TaskFieldHistoryConcurrentWriteTests.test_two_writers_in_same_transaction_append_both_transitions`  
  → 1 passed

- Task field-history backend test module passed locally  
  `docker compose -f docker-compose.dev.yml exec backend python manage.py test task.tests.test_field_history`  
  → 1 passed

- Task backend regression suite completed locally  
  `docker compose -f docker-compose.dev.yml exec backend python manage.py test task.tests --noinput`  
  → 241 total  
  → 5 skipped  
  → 0 failed

- Full Django backend suite completed locally  
  `docker compose -f docker-compose.dev.yml exec backend python manage.py test`  
  → 4349 total  
  → 21 skipped  
  → 0 failed

- Frontend ESLint verification completed  
  `docker compose -f docker-compose.dev.yml exec frontend npx eslint src/lib/taskStore.ts src/components/tasks/ListView.tsx src/components/tasks/detail/TaskFieldHistoryBlock.tsx e2e/tasks/task-history-concurrent.spec.ts`  
  → 0 errors  
  → 1 warning

- Frontend Jest regression suite completed  
  `docker compose -f docker-compose.dev.yml exec frontend npm test -- --runInBand`  
  → 127 of 128 suites passed  
  → 1 skipped  
  → 905 of 907 tests passed  
  → 2 skipped

- Two-tab Playwright E2E test passed locally  
  `npm.cmd run test:e2e -- --project=chromium e2e/tasks/task-history-concurrent.spec.ts --workers=1 --reporter=line`  
  → 2 passed

## Notes

Full-stack change  
No database migration changes  
Real-time CRDT field merging remains out of scope


quanwangniuniu (changes requested):
Scope: backend/task/signals.py and backend/task/models.py
``` 
I'm concerned about how "pending transition batches" are stored and used across the pre_save -> post-save flow.

- Where are pending batches kept while the save is in-flight (module global, threadlocal, attached to the model instance)?
- In a multi-process or threaded Django deployment this storage approach can leak between requests or lose batches if a worker crashes. Can we avoid in-memory cross-request state and instead:
  1) attach pending transitions to the Task instance (self._pending_field_transitions) and use transaction.on_commit() to append them after commit, or
  2) write history rows as part of the same DB transaction with INSERTs that reference the Task PK (no intermediate in-memory accumulation), or
  3) if there must be buffering, keep it strictly per-request (request-local) and fail fast if the buffer can't be persisted?

Please show where pending batches are held and consider changing to a per-instance or DB-backed approach (or transaction.on_commit) to avoid cross-request contamination and lost batches.
```
Why this matters: if pending batches are stored in unsafe in-memory places, concurrent requests or worker restarts can cause lost/incorrect history or cross-request contamination, breaking auditability.

Scope: backend/task/models.py (rebase + skip-no-op logic)
``` 
The "rebase each transition against the latest recorded history value" + "skip transitions that become no-ops after rebasing" needs clarification and hardening.

- How does the rebase determine ordering and the “latest recorded history value”? Is it reading history rows inside the same transaction, or querying current DB state that may have changed concurrently?
- Using timestamps for ordering is fragile (precision, clock skew). Skipping no-ops silently may drop meaningful intermediate edits (A -> B -> A) that are important for audit/forensics.
- Please document the rebase algorithm and add one of these safer changes:
  1) Make history append-only with an explicit sequence or server-generated UUID per history row and a stable DB-side ordering (e.g., SERIAL/sequence or created_at with monotonic DB-side default), and always INSERT the transition but mark it as no-op with metadata if it truly cancels out.
  2) If you must skip, persist the original attempted-before/after values (attempted_old_value, attempted_new_value) alongside the rebased result so we never lose intent.
  3) Use SELECT ... FOR UPDATE or optimistic versioning (row version) to ensure the rebasing is based on a consistent snapshot and avoid lost updates.
```
Why this matters: incorrect rebasing or silently dropped transitions breaks the historical chain and can make audits and debugging impossible; using DB-side sequencing and preserving intent keeps the history reliable.

Scope: frontend/src/lib/taskStore.ts, frontend ListView/TaskFieldHistoryBlock, and e2e test
``` 
Front-end reconciliation and history row identifiers look brittle and may cause UI loss or test flakiness.

- How are stable history row identifiers generated? If they rely on timestamps or client-side heuristics they can collide or reorder under clock skew or concurrent writes.
- How does resolveTaskFromServer reconcile out-of-order PATCH responses or partial failures? Does it rely on value-equality only (which can drop transitions if a later authoritative value matches an optimistic one)?
- Suggest returning server-assigned history entry IDs and the full resulting history in the PATCH response. On the client:
  - match optimistic edits to server responses using an operation ID (client-generated) returned by the server, not by value equality;
  - use the server history IDs to render rows and to de-duplicate or re-order deterministically;
  - when a transition becomes a no-op after rebasing, surface that as "no-op (rebased)" instead of silently dropping it, or at least log it to a client-side debug channel used in tests.

Please update the client-store reconciliation to prefer server-generated IDs and operation IDs, and update the E2E to assert on server-assigned history IDs instead of inferred timestamps.
```
Why this matters: without server-assigned stable IDs and explicit op-matching, the UI can hide edits, tests can flake under timing/clock differences, and users may lose visibility into who made which change and when.

Driiizzleeeee:
Thanks for raising these concerns.

## Change Request 1

Pending field-history batches are stored on the individual `Task` instance:

```python
instance._pending_task_field_history_batches
```

They are not stored in a module-global dictionary or a thread-local History buffer.

The synchronous `pre_save` receiver appends the transitions and author to that instance. The synchronous `post_save` receiver consumes the batch immediately and calls:

```python
TaskFieldHistory.append_transitions()
```

`Task.save()` wraps the Task write and the `post_save` History inserts in the same `transaction.atomic()` block. Therefore, if the History insert raises an exception, the Task update is rolled back as well. If the worker or database connection fails before commit, both uncommitted writes are rolled back.

I kept the instance-scoped approach rather than using `transaction.on_commit()`. An `on_commit()` callback would run after the Task update had already committed, so a callback or worker failure could leave a committed Task update without its corresponding audit History.

This keeps the pending state scoped to the individual model instance and avoids shared cross-request buffering.

## Change Request 2

### Consistent rebase snapshot

The `pre_save` receiver now locks the relevant Task row before capturing the transition:

```python
old = queryset.select_for_update().get(pk=instance.pk)
```

This runs inside the existing `transaction.atomic()` block opened by `Task.save()`. The Task-row lock is therefore held through:

```text
pre_save capture
→ Task UPDATE
→ post_save rebase
→ History INSERT
→ transaction commit
```

Concurrent writes to the same Task are serialized against the same locked Task row. The later writer performs its rebase after the earlier transaction has completed.

I selected the suggested Task-row `SELECT ... FOR UPDATE` approach. No new version or sequence field was added, so this does not require a migration.

### Rebase algorithm

The rebase runs from the synchronous `post_save` receiver inside the same Task transaction.

For each field, `append_transitions()` reads the latest persisted History row using:

```python
.order_by("-changed_at", "-pk")
```

That row’s `new_value` becomes the next row’s `old_value`.

For example:

```text
captured:
A → B
A → C

persisted:
A → B
B → C
```

The algorithm is now documented in `TaskFieldHistory.append_transitions()`.

### Stable ordering

History ordering does not depend on timestamps alone. It uses `changed_at` with the database-generated primary key as a deterministic tie-breaker:

```python
.order_by("-changed_at", "-pk")
```

The append method also compares the proposed server timestamp with the latest persisted timestamp. When necessary, it advances the next timestamp by one microsecond so that the appended transitions remain correctly ordered.

A dedicated database sequence was not introduced because that would require a model and migration change, while this ticket specifies:

```text
Migrations: no
```

### Accepted no-op transitions

Rebased no-op transitions are no longer silently skipped.

The previous logic:

```python
if old_value == new_value:
    continue
```

has been removed.

For example, if two accepted writers submit the same target value:

```text
writer 1: A → B
writer 2 captured: A → B
writer 2 rebased: B → B
```

both History rows are now persisted:

```text
A → B
B → B
```

Explicitly attempted fields are identified from `update_fields` for model saves and from the serializer’s validated update fields for API PATCH requests. This allows an accepted same-value write to remain visible without creating no-op History rows for unrelated fields.

The History UI displays these rows as:

```text
submitted [field] with no effective change
```

An actual reversal such as:

```text
A → B → A
```

also continues to produce two separate History rows.

## Change Request 3

### Stable History row identifiers

Stable History row identifiers are generated by the database through Django’s implicit:

```text
TaskFieldHistory.id
```

The ID is returned by `TaskFieldHistorySerializer`.

The UI uses the database-assigned ID for both React identity and test selection:

```tsx
<HistoryRow key={entry.id} entry={entry} />
```

```tsx
data-history-entry-id={entry.id}
```

The Playwright test also locates each History row using:

```text
[data-history-entry-id="${entry.id}"]
```

Timestamps are used for ordering and display, not as History row identifiers.

### Operation-ID reconciliation

The inline-edit flow now generates a client operation ID using:

```javascript
globalThis.crypto.randomUUID()
```

The operation ID is sent with the PATCH request as an optional query parameter:

```text
?operation_id=<client-generated-id>
```

The backend validates the ID and echoes it in the successful PATCH response. It is request metadata only and is not stored in the database, so no migration is required.

The client store tracks the latest operation using:

```text
Task ID + operation scope
```

The operation scope is based on the fields affected by that edit. Therefore, a Summary edit does not incorrectly invalidate a concurrent Priority edit on the same Task.

A server response is applied only when its returned operation ID still matches the latest operation for that Task and scope.

This protects both cases:

```text
stale successful response
→ does not overwrite the newer optimistic/server state
```

```text
stale failed request
→ does not roll back a newer successful or optimistic edit
```

Reconciliation applies only the fields owned by the operation rather than merging the entire returned Task over unrelated optimistic fields. This logic does not rely on value equality.

### History response

History continues to be retrieved through the existing authoritative field-history endpoint:

```text
GET /api/tasks/{task}/field-history/
```

That endpoint already returns authoritative History rows with their database-assigned IDs.

The client does not maintain an optimistic History collection, so there are no optimistic History rows that need value-based de-duplication. For that reason, I did not duplicate the full History collection in every PATCH response.

### Test coverage

The updated coverage includes:

* A backend test for two writers in the same transaction.
* A backend test confirming two accepted same-target writes produce `A → B` and `B → B`.
* A backend API test confirming the operation ID is echoed and an explicitly submitted no-op is recorded.
* Jest coverage for stale successful responses.
* Jest coverage for stale failure rollbacks.
* Jest coverage confirming different Task fields are tracked independently.
* Playwright coverage confirming both two-tab PATCH requests return their matching operation IDs.
* Playwright assertions based on database-assigned History row IDs.
* History UI assertions for the recorded author on each row.

quanwangniuniu (approved):
Good job

PR #778, opened 2026-08-10
feat(chat): pinned messages per channel and reliable delivery under load + MED-278

zizhenliu0427:
## Summary

**MED-278 Part 1 — chat pinned messages per channel**

[MED-278 — Chat pinned messages per channel](https://linear.app/mediajira/issue/MED-278/chat-pinned-messages-per-channel)

Completes MED-278 — channel-admin pinned messages. Channel managers can pin and unpin messages from the message menu, all members can read the pinned drawer, the list is ordered newest pin first, and pin changes now reach other members' open sessions live.

This is not a from-scratch build. The `PinnedMessage` table, the three REST endpoints, the message-menu action and the drawer section already existed on `prod-preview`. The write path was broken and neither of the required test suites existed, so this PR fixes the broken links, tightens the permission and lifecycle boundaries, adds live propagation, and adds the pytest and Playwright coverage the acceptance criteria call for.

Part 2 of this ticket (reliable delivery under concurrent load) is described separately in following comment. Both parts ship on this one branch, so a few files appear in both documents.

### Correctness and permissions

- **Pin/unpin write path restored** — `pinMessage`/`unpinMessage` sent the numeric `chat.id` into a slug-only route, so every pin click returned 404. Both now send `chat.slug`.
- **Channel-admin gate enforced** — the previous condition `chat.type == GROUP and not is_manager` fell through for private chats, letting any DM participant pin. Pin and unpin are now group-only and manager-only.
- **Input validation** — `message_id` goes through a serializer (`IntegerField(min_value=1)`) on both pin and unpin, so malformed identifiers return 400 instead of surfacing an ORM error as 500.
- **Message lifecycle** — revoked messages can no longer be pinned, deleted/revoked messages are filtered out of the list, and deleting or revoking a message removes its pin row.
- **Deterministic ordering** — the queryset orders explicitly by `-created_at, -id`, so newest pin first is stable and independent of message send time.
- **Legacy DM pins cleaned up** — tightening the permission left pre-existing private-chat pins listed but unremovable. The list endpoint now returns an empty list for non-group chats and a data migration deletes the orphaned rows.
- **Shared manager rule** — `isChannelManager` mirrors `ChatService.is_channel_manager` (explicit manager → creator → earliest active participant fallback) and is used by the timeline, thread panel and drawer, replacing three divergent inline checks.

### Live propagation

- **Pin broadcast over WebSocket** — pinning or unpinning now emits a `pin_update` event, so other members' open sessions update without switching channels or reloading. The broadcast is queued rather than sent inline on the request path (`notify_pin_update`), so a slow or unreachable channel layer cannot make the pin request itself slow or fail.
- **Unseen-pin alerts in the sidebar** — a channel that receives a pin while the user is looking elsewhere shows an indicator on its sidebar row, cleared when the user opens the channel's pinned view.

### Interface

- **Drawer states** — loading, error with retry, and empty are now distinguishable; a failed pins request previously rendered as "No pinned messages yet".
- **Pin audit context** — the drawer shows who pinned the message and when, which is the ticket's `pinned_by` / `pinned_at` surface, with relative timestamps.
- **Jump to an unloaded pin** — the drawer's jump routed through a CustomEvent that only scanned rendered messages, so clicking a pin above the loaded window did nothing. It now routes through the URL, reusing the deep-link path that pages in history.
- **Banner and drawer redesign** — the pinned banner is a single line with the pin chip, a New indicator for unseen pins, and a fade transition between pins rather than an abrupt swap. Unpin is an explicit labelled control instead of a hover-only affordance.
- **Removal animation** — unpinning collapses the row rather than making it disappear under the cursor, and the timeline no longer jumps when a pin marker is removed.
- **Mobile drawer** — the channel details panel was `hidden … md:flex`, so the Details button did nothing on small screens.
- **Duplicate-click protection** — an in-flight pin request for the same message is ignored.

## Data model note

The ticket scope lists `backend/chat/models.py — is_pinned + pinned_at + pinned_by`. Those fields are **not** added.

A dedicated `PinnedMessage` table has existed since the initial chat schema and already expresses all three:

| Ticket field | Existing equivalent |
|---|---|
| `is_pinned` | presence of a `PinnedMessage` row (unique on `chat, message`) |
| `pinned_at` | `PinnedMessage.created_at` (indexed `chat, -created_at`) |
| `pinned_by` | `PinnedMessage.pinned_by` (`SET_NULL`) |

Adding the three columns would create two sources of truth requiring double writes on pin, unpin, delete and revoke, with no behavioural gain. `makemigrations chat --check --dry-run` reports **No changes detected**, so there is no schema drift.

The ticket's `Migrations: yes` is satisfied by a **data** migration (`0013_remove_non_group_pinned_messages`) that removes the unreachable legacy DM pins described above. No schema migration is generated to satisfy the ticket text. (`models.py` does change on this branch, but for Part 2's delivery work, not for pinning.)

## Permission matrix

| Operation | Channel manager | Active member | Inactive member | Outsider | Unauthenticated |
|---|---|---|---|---|---|
| List pins | Allow | Allow | 404 | 404 | 401 |
| Pin | Allow | 403 | 404 | 404 | 401 |
| Unpin | Allow | 403 | 404 | 404 | 401 |

Inactive participants and outsiders are filtered out by `get_queryset()` before the explicit check, so they receive 404 and cannot probe channel existence. Private chats receive 403 on write and an empty list on read.

## Files Changed

Pinning accounts for roughly 40 of the branch's 88 changed files. `views.py`, `serializers.py`, `tasks.py` and `consumers.py` are also touched by Part 2.

**Backend**

- `backend/chat/views.py` — group-only + manager-only gate, serializer validation, revoked-message rejection, explicit ordering, non-group list filter, pin cleanup on message delete and revoke
- `backend/chat/serializers.py` — `PinnedMessageContentSerializer` (6-field message summary, replaces the nested full `MessageSerializer`) and `PinMessageRequestSerializer`
- `backend/chat/tasks.py` — `notify_pin_update`, the queued pin/unpin broadcast
- `backend/chat/consumers.py` — `pin_update` handler
- `backend/chat/migrations/0013_remove_non_group_pinned_messages.py` — data migration removing legacy non-group pins
- `backend/chat/tests/test_pinned_messages.py` — new suite

**Frontend — pinned surfaces**

- `frontend/src/components/chat/PinnedMessageBanner.tsx` — single-line banner, pin chip, unseen indicator, fade between pins
- `frontend/src/components/chat/PinnedMessagesDrawer.tsx` — relative pin times, explicit Unpin control, collapse-on-remove, attachment indicator
- `frontend/src/components/chat/MessageFlagChips.tsx` — unpin state machine and row collapse so the timeline does not jump
- `frontend/src/components/chat/ChannelDetailsDrawer.tsx` — loading/error/retry states, pin time, defensive sort, shared manager rule, test ids
- `frontend/src/components/chat/ChatWindow.tsx` — slug calls, in-flight dedup, pins load error toast, mobile drawer container, drawer jump routed through the URL
- `frontend/src/components/chat/ThreadPanel.tsx`, `MessageItem.tsx`, `MessageList.tsx`, `SavedItemsPanel.tsx` — pin markers and test ids
- `frontend/src/components/messages/LeftSidebar/SidebarChatRow.tsx`, `MessagePageContent.tsx` — unseen-pin indicator
- `frontend/src/hooks/useAnimatedPanel.ts`, `src/components/chat/avatarColor.ts`, `src/app/globals.css` — shared panel motion and styling

**Frontend — state and contract**

- `frontend/src/lib/api/chatApi.ts` — pin/unpin take `chatSlug`; response type narrowed to the lightweight message summary
- `frontend/src/lib/chatPermissions.ts` — shared `isChannelManager` mirroring the backend rule
- `frontend/src/lib/chatStore.ts`, `src/hooks/useChatWebSocket.ts`, `src/types/chat.ts` — `pin_update` handling and unseen-pin state
- `frontend/src/lib/messages/messagesRoutes.ts` — deep-link route used by the drawer jump
- `openapi/openapi_spec/chat.yaml` — the three endpoints, schemas, status codes, manager-only write and newest-first ordering

**Tests**

- `frontend/e2e/messages/messages-pinned.spec.ts` — new spec
- `frontend/src/__tests__/lib/chatPermissions.test.ts`, `__tests__/lib/messagesRoutes.test.ts`, `__tests__/components/SidebarChatRow.pinAlert.test.tsx`, `__tests__/hooks/useChatWebSocket.pinAlerts.test.tsx` — new unit tests
- `frontend/e2e/messages/messages-helpers.ts` — pins route matches slugs; stale `project-storage` key corrected to `project-storage-v1` to match `projectStore.ts`; onboarding-status mock and active-project cookie added so the mocked messages pages render
- `frontend/e2e/messages/messages-message-actions.spec.ts` — pins mock returns an array instead of sharing the paginated files/scheduled response
- `frontend/e2e/messages/messages-reconnect-attachments.spec.ts` — removes the now-duplicated local pins mock
- `frontend/playwright.config.ts` — registers the new spec in the mocked messages project

The `messages-helpers.ts` changes are shared by every messages spec. They are fixture corrections required to make the mocked messages pages render at all; without them the new spec cannot reach the chat window.

## Expected Result

- A channel manager can pin from the message menu; the message shows a pinned marker immediately.
- A regular member sees no Pin action and no Unpin control, but can read the pinned drawer.
- A member with the channel already open sees a new pin appear without reloading; a member looking at another channel sees an indicator on that channel's sidebar row.
- The drawer lists newest pin first, by pin time rather than message send time.
- Each entry shows the sender, a preview, who pinned it and when.
- Clicking an entry jumps to the message, including one far above the loaded window; a pinned thread reply opens its thread.
- Unpinning from the drawer removes both the drawer entry and the timeline marker, and the timeline does not jump.
- Deleting or revoking a pinned message removes it from the drawer.

## Test

Re-verified after merging `prod-preview` into this branch:

- [x] `pytest chat/tests` (full chat suite) — **208 passed**
- [x] `python manage.py makemigrations chat --check --dry-run` — **No changes detected**
- [x] `python manage.py migrate` — `0013` and `0014` applied, plus `access_control.0003` from the merge

Measured before the `prod-preview` merge, unchanged by it (frontend-only, and the merge touched unrelated frontend areas):

- [x] `pytest chat/tests/test_pinned_messages.py` — **23 passed**
- [x] Playwright `messages-pinned.spec.ts` — **8 passed** (mocked project and Chromium project)
- [x] Playwright `e2e/messages` full Chromium run — **50 passed**, 4 failed, 6 did not run
- [x] Jest `chatPermissions.test.ts` — **5 passed**
- [x] `tsc --noEmit` — 9 errors, all pre-existing in unrelated e2e specs; **0 in changed files**
- [x] `next lint` on changed files — warnings only, all on lines this branch does not touch
- [ ] Frontend Jest and Playwright re-run after the merge
- [ ] Frontend production build
- [x] Manual QA video
- [ ] PR CI passes on `prod-preview`

### Note on the 4 Playwright failures

`messages-message-actions.spec.ts:387`, `messages-reconnect-attachments.spec.ts:184`, `messages.spec.ts:172` and `messages.spec.ts:531` fail on this branch. They were verified as pre-existing by checking out the `prod-preview` versions of the affected spec and helper files and reproducing the identical failure. The count is unchanged from the pre-change baseline (49 passed / 4 failed before, 50 passed / 4 failed after, the extra pass being the new jump test).

### Note on the jump regression test

`jumping to a pin above the loaded window pages the message in` was confirmed to fail against the previous CustomEvent implementation, failing on the jump assertion rather than during setup, so it genuinely covers the fix.

## Out of Scope / Follow-up

- **Pin ordering by user** — excluded by the ticket.
- **Pins pagination** — the list returns a plain array. The nested serializer was reduced to a 6-field summary and a test asserts the query count is identical for 2 and 10 pins, so there is no N+1. Pagination would change the response contract and is not requested.
- **Numeric chat ids on other chat endpoints** — `updateChatDetails`, `updateNotificationSettings`, `addParticipant`, `removeParticipant`, `updateParticipantManager`, `listChatFiles`, `leaveChat` and `deleteChat` still send `chat.id` into slug-only routes and return 404. This is unfinished MED-231 work, unrelated to pinning, and is left untouched here; it needs its own ticket. Note that "Make manager" in the details drawer is affected, so QA should demo with a channel the manager account created.

Ticket: [MED-278](https://linear.app/mediajira/issue/MED-278/chat-pinned-messages-per-channel)

Following is MED-278 Part 2


zizhenliu0427:
## Summary

**MED-278 Part 2 — reliable chat message delivery under concurrent load**

Resolves the message failure discovered during MED-278 QA: at approximately 100 simultaneous chat users, sends became very slow, PostgreSQL exhausted its connection slots, the page/server appeared frozen and some clients did not receive every message.

Before the code change, I noticed prod-preview already used Celery (Message Queue) with Redis, but only part of the send pipeline was asynchronous. This change makes the database-to-queue hand-off durable, removes O(recipients) status expansion from the HTTP hot path, isolates chat queues, restores tenant context in background paths, routes application database traffic through a safe PgBouncer configuration, and closes two channel-layer defects that were discarding messages with no error and no log.

**No new message queue/broker is introduced.** Kafka and RabbitMQ are out of scope; see "Why not Kafka or RabbitMQ" below.

Part 1 of this ticket (pinned messages per channel) is described separately in `MED-278-PR_DESCRIPTION.md`. Both parts ship on this one branch, so a few files appear in both documents.

### Durability of the send path

- **Transactional outbox** — `ChatOutboxEvent` is committed with the message. A public-schema dispatcher claims with `skip_locked`, publishes to Celery, marks successful hand-offs and releases/retries failed claims with backoff.
- **O(1) HTTP write path** — a large-group send no longer creates every recipient status in the request. `notify_new_message` creates missing statuses idempotently in the worker.
- **Idempotent client retries** — `client_message_id` is unique per sender. Retried or concurrent requests return the committed message without duplicate side effects.
- **One message-creation path** — legacy WebSocket `chat_message` creation is rejected; REST persistence plus idempotency and outbox is authoritative.
- **Scheduled-message parity** — scheduled sends create statuses and durable outbox rows before marking the scheduled item sent.
- **One claim mechanism** — the realtime fan-out, the offline delivery task and reconnect recovery all take ownership by winning the same `sent -> delivered` row transition, so no two paths publish the same message to the same recipient.

### Channel-layer message loss

Two separate silent-drop mechanisms were found by load testing. Neither raised, and only one of them logged.

- **Capacity** — `channels_redis` defaults `capacity` to 100 per channel, and every WebSocket connection served by one ASGI process shares a single inbound queue. When `group_send` finds the queue full it skips the write, logs one INFO line, and raises nothing. Measured with 100 users in one channel: capacity 100 → 15,471 drops logged and 78 delivered; capacity 50,000 → 0 drops. Now set explicitly.
- **Expiry** — this one has no log at all. Every `group_send` sweeps each recipient channel with `ZREMRANGEBYSCORE(key, 0, now - expiry)` before writing, so anything a connection has not read inside the window is deleted. The default is 60s. Driving ~1,700 deliveries/s against a single-process ceiling of roughly 1,000/s, the backlog grew past the window and **~12,000 of 148,500 deliveries (8%) vanished** — while every message was committed, every outbox event published, every socket stayed open, and `MessageStatus` recorded them all as delivered. `expiry` is now explicit at 300s, and the same run delivers **148,500 of 148,500**.

### Delivery is now recoverable, not just recorded

- **Reconnect replay** — a message is marked delivered when it is handed to the channel layer, not when a socket writes it, and reconnect recovery only looked at rows still marked `sent`. Anything lost after the hand-off was therefore unrecoverable. Reconnect now also replays messages marked delivered inside the expiry window; the client discards what it already holds by message id, so replaying more than necessary is the cheap side of the trade. This covers every way a publish can be accepted and then lost, not only expiry.

### Connections and tenancy

- **Tenant-safe background work** — a shared validated schema context wraps Celery, WebSocket, agent and scheduled-message ORM operations and restores the previous schema afterwards.
- **Database admission** — a process-local semaphore bounds active database request chains and returns a retryable 503 rather than exhausting PostgreSQL.
- **PgBouncer actually enabled** — Compose now uses Django's real `POSTGRES_PORT` setting and routes backend/workers through the `pgbouncer` service. Connection setup fell from 50.88 ms to 2.96 ms, which is what removed the connect-time queueing behind daphne's handshake timeout.
- **Tenant-compatible pooling** — session pooling plus `DISCARD ALL` is used because tenant selection relies on session-level `SET search_path`; transaction pooling is not safe for that application contract.
- **PgBouncer auth pinned** — `AUTH_TYPE` is set explicitly per environment instead of relying on the image default, which silently failed against SCRAM-SHA-256 password hashes.
- **SSE no longer pins a connection** — the notification stream held a PostgreSQL connection open for the life of the stream. Released inside the generator, because the tenant middleware's `finally` block runs the moment the view returns the response object.
- **WebSocket handshake headroom** — daphne's `--websocket_connect_timeout` defaults to 5s, which a 100-connection burst exceeded while queueing. Now configurable, default 30s. nginx `worker_processes`/`worker_connections` and the `/ws/` timeouts were corrected to match.

### Fan-out

- **Chat-group broadcast, behind a flag** — one channel-layer publish per chat instead of one per recipient. Implemented and verified, but **`CHAT_CHANNEL_GROUPS_ENABLED` defaults to `False`, so this PR does not change production fan-out behaviour.** It changes where the authorisation decision is made (group membership rather than per-recipient targeting), which is why it is not enabled by default in the same PR that introduces it. Enabling it is an env-var change, not a code change.

Both paths were measured at 100 users on the same host. Both deliver all 9,900; the flag only changes latency, so leaving it off costs headroom, not correctness. Figures are in the Test section, labelled by flag state.

## Why not Kafka or RabbitMQ

The reported symptom — messages slow, some never arriving — reads like a missing queue, so this was checked before any design work. It is not what happened.

Celery on Redis was already in place on `prod-preview` and is the right tool here. The defect was that the pipeline was only partly asynchronous and the database-commit-to-broker boundary was not durable: a committed message whose `.delay()` was rejected left nothing behind that guaranteed the work would ever run. A second broker does not fix that; a transactional outbox does, which is what this PR adds. Replacing the broker would have been a large migration addressing a problem that was never the cause.

The audit also found Kafka provisioned in the Docker stack but carrying no application traffic: no `KafkaProducer`, `KafkaConsumer` or `kafka` import exists anywhere in the backend — `KAFKA_BROKER`, `KAFKA_ENABLED` and `KAFKA_TOPICS` are settings definitions with no call site — and the running broker reported no consumer groups. So Kafka was neither the existing chat queue nor a candidate explanation for the backlog.

Kafka is left untouched, because "unused" and "not yet used" look identical from the outside and I cannot tell them apart from the code alone.

> **Question for reviewers:** is the Kafka stack empty because another project is actively building on it, or is it left over? If someone is depending on it, this PR does not touch it and nothing here needs to change. If it is genuinely abandoned, removing it should be its own ticket — it holds roughly 1.8 GB across broker, UI and exporter in the local stack, and every developer running the dev Compose file pays that.

## Architecture

```text
Browser
  │ REST POST + client_message_id
  ▼
Django admission limit
  │
  ▼
PostgreSQL transaction
  ├─ tenant Message
  └─ public ChatOutboxEvent(s)
          │
          ▼
Celery Beat dispatcher
  ├─ chat.realtime ── statuses + WebSocket publish
  │                    (per recipient by default;
  │                     one chat-group publish when the flag is on)
  ├─ chat.notifications
  └─ chat.delivery ── offline/retry delivery
```

PostgreSQL connections from Django and Celery pass through PgBouncer. Django uses `CONN_MAX_AGE=0`; after a request/client session ends, the server connection returns to the session pool and is reset with `DISCARD ALL` before another tenant can use it.

## Reliability guarantees

| Failure point | Behaviour |
|---|---|
| Client retries after timeout | Same sender + `client_message_id` returns the existing message |
| Redis unavailable after DB commit | Unpublished outbox row remains durable and is retried |
| Dispatcher dies after claiming | Stale claim expires and another dispatcher can reclaim it |
| Realtime publish fails | Recipient claims are released for delivery retry |
| Publish succeeds but the queued message expires | Reconnect replays it from the delivered-recently window |
| User is offline | `MessageStatus` remains durable and reconnect recovery delivers it |
| Worker runs outside HTTP middleware | Task enters the recorded tenant schema explicitly |
| 100 HTTP requests arrive together | Admission limit and PgBouncer preserve database connection capacity |

## Key Files Changed

**Message persistence and delivery**

- `backend/chat/models.py` — sender-scoped `client_message_id` and `ChatOutboxEvent`
- `backend/chat/services.py` — idempotent persistence, durable outbox creation, deferred recipient status fan-out, the shared delivery claim
- `backend/chat/tasks.py` — outbox dispatcher, tenant-aware realtime/status creation, group broadcast and scheduled-message parity
- `backend/chat/consumers.py` — direct-send rejection, chat-group membership, single-writer send with per-connection dedup, reconnect recovery and delivered-recently replay
- `backend/chat/realtime.py`, `backend/chat/metrics.py` — broadcast helper and enqueue-failure counter
- `backend/chat/views.py` / `serializers.py` — REST idempotency plumbing
- `backend/chat/migrations/0012_message_client_message_id.py` / `0014_chatoutboxevent_and_more.py` — durable schema changes

**Channel layer and connection safety**

- `backend/backend/settings.py` — channel-layer `capacity` and `expiry`, reconnect replay window, middleware, Celery routing, outbox schedule, connection labels and the chat-group flag
- `backend/core/tenant_context.py` — validated tenant schema context
- `backend/core/middleware/database_concurrency.py` — bounded request admission
- `backend/notifications/sse.py`, `views.py` — connection release inside the stream generator
- `backend/agent/tasks.py` / `services.py` — tenant-aware agent replies through the same message service
- `backend/entrypoint`, `entrypoint-dev`, `Dockerfile*` — configurable WebSocket handshake timeout
- `nginx/nginx.conf`, `.preview`, `.pro` — worker sizing and `/ws/` timeouts

**Deployment and load verification**

- `docker-compose.dev.yml`, `.preview.yml`, `.pro.yml`, `.pro.single.yml`, `.pro.dr.yml` — PgBouncer with pinned auth type plus independent chat workers
- `postgres/postgresql.conf` — connection sizing
- `env.example` — channel-layer, replay, connection and admission configuration documentation
- `k6/scripts/scenarios/` — realtime load test with an enforced send barrier, plus ordering, dual-connection, jitter, reconnection-storm, soak and spike scenarios
- `backend/chat/management/commands/prepare_chat_load_test.py` — distinct 100-user QA credentials and clean group preparation
- `backend/chat/tests/`, `backend/core/tests/test_database_concurrency_middleware.py`, `backend/notifications/tests/test_sse.py` — reliability regression coverage

## Expected Result

- A committed message is eventually handed to Celery even if the broker is temporarily unavailable.
- Retrying the same client send does not create another message or another outbox event.
- A 100-member channel can accept 100 sends without PostgreSQL connection-slot 500s.
- Every online peer receives every message once: 100 senders × 99 peers = 9,900 deliveries.
- A backlog does not turn into silent loss until it exceeds five minutes, and a message lost after being recorded as delivered is recovered on reconnect.
- Realtime work is not blocked behind unrelated default-queue tasks.
- Offline users recover durable pending messages after reconnect.
- Scheduled and agent-generated messages follow the same tenant/outbox path.
- No tenant schema state leaks when PgBouncer reuses server connections.

## Test

**Configuration and tenancy**

- [x] Five Compose variants — `docker compose config -q`
- [x] Runtime application DB — host `pgbouncer`, port `5432`, `CONN_MAX_AGE=0`
- [x] PgBouncer runtime config — session mode, wildcard mapping, `DISCARD ALL`
- [x] Tenant reuse probe — **0/20 mismatches**; **20/20** fresh sessions reset to `public`
- [x] Backend health — HTTP 200; `manage.py check` reports 0 issues

**Automated suites** (re-run after merging `prod-preview`)

- [x] `pytest chat/tests` — **208 passed**
- [x] `python manage.py migrate` — chat `0012`/`0014` and `access_control.0003` applied

### Hosts

Latency is strongly host-dependent, so every figure below names the machine it came from. Reliability counts (checks, deliveries, duplicates, failures) pass on all three and are not averaged.

| Tag | Machine |
|---|---|
| **Xeon** | MacPro5,1, dual Intel Xeon X5690 12-core/24-thread, 64 GB, macOS 15.7.1 — where the channel-layer defects were found and fixed |
| **M4 Pro** | MacBook Pro (Mac16,8), Apple M4 Pro 14-core, 48 GB, macOS 26.5.2 — the controlled same-machine before/after |
| **9800X3D** | AMD Ryzen 7 9800X3D 8-core/16-thread, 96 GB, Windows 11 Pro 25H2 + Docker Desktop — the original QA host where the defect was first reproduced |

**100-user delivery — shipping default (chat-group flag OFF) — 9800X3D**

This is the configuration this PR actually ships.

- [x] 100-user single-group burst — **301/301 checks**, 100/100 HTTP, 100/100 WS, **9,900/9,900 deliveries**, 0 duplicates, 0 send/connect failures, 0/102 HTTP failures
- [x] Latency at that run — HTTP p95 **9.15 s**, WS delivery p95 **11.39 s**, both inside the 30 s reliability gates
- [x] 100-user DB verification — 100 messages, 9,900 statuses all delivered, 200/200 outbox events published
- [x] Post-run drain — pending outbox 0 and all four Celery queues 0

**100-user delivery — chat-group flag ON — 9800X3D**

Recorded because it is the same code with one env var flipped, and it is what a follow-up would enable.

- [x] Enforced barrier, three consecutive runs — **9,900/9,900** each, 0 duplicates, WS delivery p95 **4.69–5.10 s** against **9.40 s** with the flag off
- [x] Handshake cost for a user in 200 chats — **68.5 ms**
- [x] Cross-process group revocation — a socket on one daphne process stops receiving a chat after removal performed from another process

The 9800X3D flag-ON WS p95 of 4.69–5.10 s and the M4 Pro's 4.96 s were produced by the same code, scenario and flag state on unrelated hardware and architectures, which is the closest thing here to an independent reproduction of the delivery-latency result.

**Same-machine before/after — Apple M4 Pro 14-core, 48 GB, 2026-08-10**

Three official runs per side, identical topology, warm-up, session length, HTTP timeout and 30 s gates. `prod-preview` at `c1a76d117` ran from a separate worktree against a disposable database. The fix branch was at `665654f3c`, where **`CHAT_CHANNEL_GROUPS_ENABLED` still defaulted to `True`** — so the fix-branch column below is the flag-ON path, not the shipping default. The shipping default was measured separately (above) and also delivers 9,900/9,900.

| Metric | `prod-preview` median | Fix branch median (flag ON) |
|---|---:|---:|
| Checks | 201/301 | **301/301** |
| Accepted message POSTs | 100/100 | 100/100 |
| Peer deliveries | 8,232–8,781 / 9,900 | **9,900/9,900** |
| Clients receiving every peer message | **0/100** | **100/100** |
| Duplicates | 0 | 0 |
| HTTP p95 | 9.10 s | **1.79 s** |
| WS delivery p95 | **47.49 s** | **4.96 s** |

`prod-preview` accepted every send and opened every socket on this host, then failed the peer-delivery check for all 100 clients and missed the WS p95 < 30 s gate. That is the same defect reported in QA, reproduced under control.

**Original dual-Xeon baseline**

The investigation started here, on `prod-preview` at `483bc1bc1`. This host showed the connection-slot failure mode rather than the M4 Pro's incomplete fan-out:

| Metric | `prod-preview` baseline | After the fix |
|---|---:|---:|
| Accepted message POSTs | 94/100 | 100/100 |
| HTTP 500 responses | **6** | **0** |
| HTTP p95 | 49.65 s | 16.49 s |
| Recipient status rows | 9,306 for 94 messages | 9,900 for 100, all delivered |
| Celery backlog after the run | **141 tasks** on the default queue | all four queues 0 |

**Sustained load — 9800X3D** (10× and beyond the reported failure, to find where it breaks rather than to certify capacity)

- [x] 1,000 messages × 99 peers — **99,000/99,000**, 0 duplicates, WS p95 **33.5 s**
- [x] Overload at ~1,700 deliveries/s, before the expiry fix — 136,500 of 148,500, **8% lost silently**, WS p95 pinned flat at **60.0 s**
- [x] Same run after the expiry fix — **148,500/148,500**, 0 duplicates, every receiver exactly 1,485, WS p95 **79 s** (the true backlog, no longer truncated)

**Not yet done**

- [ ] Preview-environment smoke
- [ ] PR CI passes

### Deployment note

`CHANNEL_LAYER_EXPIRY` takes effect on the **publishing** side — the expiry sweep runs inside `group_send`, which executes in the Celery chat workers, not in the web process. Restarting only the backend leaves the old 60s window in force. This was observed directly: with the backend alone restarted, the overload run still lost 5%; restarting the chat workers took it to zero.

### Capacity note

A single ASGI process sustains roughly **1,000 deliveries/s**, which for a 100-member channel is about 10 messages/s in that channel. Beyond that a backlog accumulates; the expiry setting decides how long that backlog can last before delivery becomes lossy, and reconnect replay decides whether the loss is recoverable. Raising throughput needs multiple ASGI processes, which is a deployment-topology change (daphne cannot fork, so it means Compose replicas or a different ASGI server) and is deliberately not in this PR. The measured headroom is roughly two orders of magnitude above current usage.

### Performance note

HTTP send p95 spans roughly an order of magnitude across the three hosts — 16.49 s on the Xeon, 9.15 s on the 9800X3D, 1.79 s on the M4 Pro — on the same code and scenario, so no single number characterises the system. Reliability does not vary that way: every host delivers 9,900/9,900 with 0 duplicates and clears the 30 s gates.

The script's strict 500 ms HTTP / 1 s WebSocket gates are **not** claimed as passed on any host. This PR claims reliability and bounded recovery under a 100-user simultaneous burst, not a production latency SLO — that needs separate capacity work on production-like hardware.

### PgBouncer isolation note

The previous PgBouncer configuration was not actually in use: Compose supplied `DB_PORT=6432` while Django reads `POSTGRES_PORT`, so application processes still connected straight to PostgreSQL. Enabling it exposed a second problem — the old pool used transaction mode, and a forced public/tenant reuse probe against that configuration produced **17 schema mismatches out of 40 clients**, because tenant selection depends on session-level `SET search_path`. Session pooling with `DISCARD ALL` is therefore required until tenant selection stops depending on session state. The same probe against the committed configuration returns 0 mismatches.

The removed static `pgbouncer/userlist.txt` contained an obsolete local-only password. It does not match the current `.env` or `env.example` and no preview/production use was identified; if reuse is found later, rotate it.

## Out of Scope / Follow-up

- **Kafka, RabbitMQ or any replacement message queue/broker** — not needed for this defect; see the section above. Removing the unused Kafka stack, if it is confirmed abandoned, belongs to its own ticket
- Multiple ASGI processes and the Compose topology change they require
- Enabling `CHAT_CHANNEL_GROUPS_ENABLED` by default, after it has run in preview
- Strict production latency SLO and capacity sizing
- WebSocket session-count metrics and Grafana dashboard (MED-319)
- Redis instance separation between Channels and Celery
- Frontend multi-window WebSocket sharing
- Long-duration soak, reconnection-storm and network-jitter certification
- Removal of the unreachable legacy `handle_chat_message` method body; the input is already rejected before it can execute

Ticket: [MED-278](https://linear.app/mediajira/issue/MED-278/chat-pinned-messages-per-channel)

quanwangniuniu (commented):
1) Concurrency / ordering and idempotency of pin_update broadcasts
- What I saw: notifications are queued rather than sent inline; this is good for request latency, but can create race conditions where multiple pin/unpin actions execute close together and broadcast order or final state observed by clients is inconsistent.
- Why fix: clients could briefly observe incorrect pin state, or stale state might overwrite a newer update. Under load this can surface as inconsistent UI (e.g., a message appears pinned after it was unpinned) which undermines correctness guarantees.
- Recommendation: ensure broadcasts are produced only after the DB transaction commits and include an ordering/sequence or version (e.g., pin.updated_at or pin.sequence) so consumers can ignore out-of-order events; add integration tests that simulate concurrent pin/unpin and assert eventual consistency. Make handlers idempotent.

2) Permission/authorization edge-cases and error semantics
- What I saw: pin/unpin restricted to group chats and channel managers. Legacy DM pins are removed via migration.
- Why fix: tightening permissions can surface unexpected 4xx responses in clients. The API should consistently return 403 for forbidden rather than 404/500 and the frontend must gracefully handle these responses (showing explanatory message, not UI errors). Also confirm the channel manager check matches all surfaces (timeline, threads, message menu) and there aren’t remaining divergent inline checks.
- Recommendation: audit all endpoints and client code paths that pin/unpin or render the unpin control; ensure the same is_channel_manager logic is used everywhere. Add tests covering non-manager attempts return 403 and the UI displays an actionable message.

3) Data-migration safety, idempotency, and rollout plan for legacy DM pins
- What I saw: a data migration deletes orphaned/legacy DM pins.
- Why fix: destructive migrations can cause data loss if applied incorrectly; migration must be idempotent and reversible (or at least safe) across environments and retries. If the migration runs during a high-traffic period, race conditions may remove rows while other operations run.
- Recommendation: include the migration in a single, tested migration file with a reversible/at-least-documented plan; ensure it runs under transactional safety. Add a smoke test that verifies pre/post migration counts and ensure DB backups or a rollback plan exist. If possible, run the cleanup in a background task with logging rather than an immediate destructive migration.

4) DB modeling: foreign keys and cascade semantics for pinned messages
- What I saw: the PR says deleting/revoking a message removes its pin row; unclear whether this is enforced by DB-level foreign keys with ON DELETE CASCADE or by application-level cleanup code.
- Why fix: application code removal can miss edge cases and be less performant, leading to orphan rows or N+1 problems. DB-level cascades are stronger guarantees.
- Recommendation: ensure PinnedMessage.message_id has a foreign key to Message with ON DELETE CASCADE (or equivalent), and add DB-level constraints and tests. If application-level deletion is required (e.g., soft deletes), document and test the cleanup path thoroughly.

5) Validation, API contract, and error handling consistency
- What I saw: message_id is validated with IntegerField(min_value=1), returning 400 for malformed identifiers rather than 500.
- Why fix: returning 400 is good, but be sure validation covers all inputs (string ids, very large ints), and that authentication/authorization errors are returned as 401/403 consistently. Also ensure API consumers (mobile/web) are updated to the contract.
- Recommendation: add unit tests for invalid payloads, missing fields, and extreme values; update API docs (OpenAPI/Swagger) with the validated schema; ensure integration tests confirm 401/403 vs 400 semantics are correct in different failure modes.

6) Frontend UX/performance and websocket handling for large channels
- What I saw: drawer improvements, jump-to-pin fixed via routing, unseen-pin indicator, and animations. Broadcast-based live updates are implemented.
- Why fix: UI correctness depends on realtime events — ensure the client correctly deduplicates/batches pin_update events, clears the unseen indicator when the drawer is opened, and handles large pin lists (pagination or lazy-loading). Also ensure deep-link jumps to messages not yet loaded perform a reliable fetch and scroll behavior. Without that, users may get stuck or see inconsistent state.
- Recommendation: add Playwright tests for:
  - receiving pin_update while on another channel and confirming the sidebar indicator appears and clears when opening the pinned view,
  - jumping to a pinned message that isn’t loaded (deep link) and verifying the message is loaded/scrolled into view,
  - behavior under rapid pin/unpin sequences to validate client debouncing/deduping. Also review accessibility (keyboard/aria) for the new drawer states and controls.

zizhenliu0427:
Thanks for the review — point 1 found a real race, though the mechanism turned out to be one layer below where it was described. That is now fixed, along with the input-bound gap in point 5. For the rest, the behaviour asked for is already in place, and there are two recommendations I would rather not take; reasons below.

---

## 1. Ordering and idempotency of `pin_update` broadcasts

**Fixed — but not where the review located it.**

Two parts of this were already true:

- Broadcasts are already commit-gated. `_notify_pin_update` wraps the enqueue in `transaction.on_commit(...)` (`backend/chat/views.py:227`), so nothing is published for a pin that never committed. Covered by `test_pin_queues_broadcast_instead_of_publishing_in_request` and `test_unpin_queues_broadcast_instead_of_publishing_in_request`.
- The handlers are already idempotent, because the event carries no state. `handleSocketPinUpdate` (`frontend/src/components/chat/ChatWindow.tsx:269`) ignores the payload and calls `refreshPins()`; the event is a cache-invalidation signal and the server remains the only source of truth. N events produce N refetches and the same final state.

That second point is why I did not add a sequence or version field: out-of-order *events* cannot produce wrong state when no event content is ever applied. "A message appears pinned after it was unpinned" is not reachable through event ordering.

The race is real one level down, in the refetch itself. `refreshPins` had no request-sequence guard, so:

- a rapid pin/unpin sequence puts several `listPins` calls in flight, and an older response resolving last would overwrite a newer list;
- switching channels did not invalidate the previous channel's in-flight request, so a late reply could render one channel's pins inside another.

Fixed with a generation counter — only the newest request may write. Switching channels re-enters `refreshPins`, which bumps the counter and invalidates the previous channel's request as a side effect.

## 2. Permission and error semantics

**Already in place, and I would push back on the 403-everywhere suggestion.**

- Non-managers already receive an explicit 403, not 404 or 500 (`backend/chat/views.py:694` for pin, `:728` for unpin).
- The 404 for non-participants and outsiders is deliberate, not an oversight. `get_queryset()` filters them out before the permission check, so they cannot distinguish "channel exists but I am not allowed" from "no such channel". Returning 403 there would confirm a channel's existence to anyone who can guess a slug, which is a channel-enumeration vector. The permission matrix in the PR description states this for each caller class.
- The manager rule is single-sourced: `isChannelManager` (`frontend/src/lib/chatPermissions.ts`) is used at `ChatWindow.tsx:59` and `ChannelDetailsDrawer.tsx:669,676`. There are no divergent inline checks left — the nearby `creator` computation in `ChannelDetailsDrawer` builds a display label, not a permission decision.
- Covered by tests for a regular member, a private-chat participant, an inactive participant, an outsider, and an unauthenticated caller.

## 3. Data-migration safety for legacy DM pins

**Already idempotent, reversible and transactional.**

`0013_remove_non_group_pinned_messages` is `RunPython(delete_non_group_pins, migrations.RunPython.noop)`:

- **Reversible** — the reverse is an explicit no-op. Deleted rows cannot be reconstructed, so a fabricated inverse would be worse than an honest one: migrating backwards succeeds and does not pretend to restore anything.
- **Idempotent** — the queryset matches nothing on a second run, so retries are safe.
- **Transactional** — `RunPython` runs inside the migration's transaction on PostgreSQL.

On the high-traffic race: the rows this deletes are unreachable by construction. The list endpoint hides non-group pins and unpin returns 403 for them, so no concurrent request can read, create or remove one while the migration runs.

I would rather not move this into a background task. It would leave an open-ended window in which the rows still exist, add an operational step to every environment, and make "has the cleanup finished?" a question someone has to answer later — in exchange for no additional safety over a transactional, idempotent, reversible migration.

## 4. Foreign keys and cascade semantics

**Enforced by the model.**

`PinnedMessage.message` and `PinnedMessage.chat` are both `on_delete=models.CASCADE` (`backend/chat/models.py:851` and `:845`), so a hard delete of either takes the pin row with it.

The application-level cleanup exists for a case a cascade cannot cover: revoking is a *soft* delete. No row is removed, so no foreign key can fire, and the pin has to be cleared explicitly. Both paths are tested — `test_deleting_message_removes_its_pin_record` and `test_revoking_message_removes_its_pin_record`.

## 5. Validation and API contract

**Covered, and the "very large ints" sub-point was a real gap — now closed.**

Already present:

- `test_pin_rejects_invalid_message_id` is parametrized over a missing field, a non-numeric string, `0` and `-1`.
- 401, 403 and 400 are asserted separately, in their own tests, so the distinction between the three is pinned rather than assumed.
- `openapi/openapi_spec/chat.yaml` carries the validated schema, status codes, manager-only write and ordering.

The gap: `message_id` was `IntegerField(min_value=1)` with no ceiling. DRF's `IntegerField` accepts any Python int, so an id larger than the column could hold passed validation and only failed once PostgreSQL rejected it — a 500 for a plainly malformed request. `Message` uses `BigAutoField`, so the bound is now `max_value=2**63 - 1`, chosen from the column type rather than picked arbitrarily, with `2**63` added to the parametrized test.

## 6. Frontend realtime handling and coverage

**Mostly covered; one genuine gap, which I am happy to close if you want it.**

- **Dedup/batching of `pin_update`** — not needed for correctness, for the reason in point 1: the event carries no state. The rapid-sequence problem was the refetch race, and the generation guard is the fix for it.
- **Clearing the unseen indicator** — `acknowledgePins` calls `clearChatPinUnseen` (`ChatWindow.tsx:212`), and `applyPins` clears it when the channel has no pins left. Unit-tested in `SidebarChatRow.pinAlert.test.tsx` and `useChatWebSocket.pinAlerts.test.tsx`.
- **Deep-link jump to an unloaded message** — this Playwright test already exists: `jumping to a pin above the loaded window pages the message in` (`frontend/e2e/messages/messages-pinned.spec.ts:500`). It was confirmed to fail against the previous CustomEvent implementation, failing on the jump assertion rather than during setup, so it covers the fix rather than passing vacuously.
- **Large pin lists / pagination** — deliberately out of scope and called out in the PR description. The response is a plain array, the nested serializer was reduced to a 6-field summary, and `test_list_pins_query_count_does_not_grow_with_pins` asserts the query count is identical for 2 and 10 pins, so there is no N+1 to pay for. Pagination would change the response contract and no acceptance criterion asks for it.
- **Not covered, and you are right that it is missing**: a Playwright test for receiving a `pin_update` while viewing a *different* channel and asserting the sidebar indicator appears and then clears. Today that path only has Jest coverage. An accessibility pass over the new drawer states and the Unpin control is also not done. Say the word and I will add both.

---

## Verification after these changes

- `pytest chat/tests/test_pinned_messages.py` — **27 passed** (23 before, plus the new bound case)
- `pytest chat/tests` — **208 passed**
- `tsc --noEmit` — 9 errors, all pre-existing in unrelated `e2e/` specs; **0 in changed files**
- Jest, pin and permission suites — **8 passed**

quanwangniuniu (changes requested):
Scope: General — background pin broadcast and ordering / transactional delivery

```` 
Comment:

The change to enqueue pin-update broadcasts off the main request path (notify_pin_update) is the right idea for availability, but I see two risks that need mitigation:

1) enqueue-before-commit / lost-notifications: if the background job is scheduled before the DB transaction that creates/deletes the pin is committed, workers may observe no pin row and skip or error. Ensure broadcasts are enqueued only after the DB commit (e.g. use transaction.on_commit or equivalent).

2) ordering & idempotency under concurrency: concurrent pin/unpin/delete operations can be processed out-of-order by the worker, causing clients to see stale or flipped state. Make broadcasts idempotent and include a deterministic ordering key (pin_version, pin_id + created_at, or a monotonic sequence) so clients can apply updates in the correct order.

Suggested changes:
- Enqueue notify_pin_update from a transaction.on_commit callback (or equivalent) so workers only see committed state.
- Include an explicit version/sequence or the pinned row's created_at + id in the event payload and have clients ignore older events.
- Make worker handling idempotent and add retries/dead-letter behavior for transient failures; add monitoring/metrics for failed notifications.
- Add unit/integration tests that simulate concurrent pin/unpin/delete and assert eventual consistent state and ordering.

Example (Django/Celery-style pseudocode):
```python
# inside view/service handling pin
with transaction.atomic():
    pin = PinnedMessage.objects.create(...)
    # schedule broadcast only after commit
    transaction.on_commit(lambda: notify_pin_update.delay(chat_slug=chat.slug, pin_id=pin.id, pin_version=pin.version))
```

````

Why this matters
- If the broadcast is queued before commit, workers may not see the created pin and notifications will be lost or inconsistent.
- Without ordering/versioning, concurrent operations can be applied out-of-order on clients, leading to confusing UX (e.g., an unpin arriving before the pin).
- These failures are high-impact because they break live collaboration guarantees; the fixes are small (on_commit + versioning + idempotency) and greatly reduce race-related bugs.

---

Scope: General — permission change + data migration that deletes legacy DM pins / isChannelManager fallback

```` 
Comment:

The tightened permission checks and the cleanup migration are correct directions, but two concerns stand out:

1) destructive migration risk: the migration "deletes the orphaned rows" for legacy DM pins — that is irreversible in production unless you have a full DB restore. Consider moving rows to an archive table or marking them as migrated/deprecated rather than deleting, and run the deletion as a separate, auditable step after verification.

2) implicit privilege escalation via manager fallback: the new isChannelManager logic falls back to "earliest active participant" when no explicit manager exists. That could unintentionally grant manager-only actions (pin/unpin) to a user who shouldn't have them, especially in older channels. Either tighten the fallback, document it clearly, or require an explicit promotion path for managers.

Suggested changes:
- Change the migration to a non-destructive transform first:
  - create pinned_message_archive and copy orphaned rows there, add metadata (migrated_at, original_chat_slug).
  - return an admin report listing affected chat IDs so product/PM can approve deletion.
  - only run hard-delete in a second, scheduled step after review.
- Adjust isChannelManager behavior:
  - prefer explicit manager roles; remove or restrict the "earliest active participant" fallback OR gate it behind a feature flag and add a migration that records which channels would be affected.
- Add automated tests covering: non-group chat pins remain unreachable from the UI; channels with no explicit manager do not suddenly grant manager actions to arbitrary users.

Example migration approach:
```sql
-- create archive
CREATE TABLE pinned_message_archive AS TABLE pinned_message WITH NO DATA;
-- copy orphaned
INSERT INTO pinned_message_archive SELECT *, now() as migrated_at FROM pinned_message pm JOIN chat c ON pm.chat_id = c.id WHERE c.type != 'group';
-- (optionally) verify counts/report, then delete in a follow-up step
```

````

Why this matters
- A destructive migration that deletes rows without an easy recovery path risks permanent data loss and support escalations. Staging the cleanup and keeping an audit/archive preserves recoverability.
- Implicit privilege fallback can create security/UX surprises (users getting manager powers unexpectedly). Small changes to the fallback policy and additional tests make the permission model predictable and safe.

---

**Comment by zizhenliu0427** [comment]
Thanks Ray. Most of the first comment is already implemented, one part of it does not apply to how the client consumes these events, and the remaining parts (metrics, tests) are now done. On the second comment I have taken the migration concern and implemented a lighter version of it; on the manager fallback I want to push back, because the fallback is not something this PR introduces and tightening it here would lock most channels out.

---

# 1. Pin broadcast ordering and transactional delivery

## on_commit — already in place

`queue_pin_update` wraps the enqueue in `transaction.on_commit`, so a worker never runs for a pin that was not committed:

```python
def queue_pin_update(chat_id, action, message_id, pin_data=None, actor_user_id=None):
    ...
    def enqueue() -> None:
        try:
            notify_pin_update.delay(...)
        except Exception:
            chat_broadcast_enqueue_failures_total.labels(event='pin').inc()
            logger.exception('Failed to queue pin update for chat %s', chat_id)

    transaction.on_commit(enqueue)
```

`backend/chat/views.py`. This is the same shape as your pseudocode. Covered by `test_pin_queues_broadcast_instead_of_publishing_in_request` and `test_unpin_queues_broadcast_instead_of_publishing_in_request`, both of which assert the task is queued through the commit hook rather than called inline.

## Versioning — does not apply, because the client never applies the payload

The handler ignores the event body and refetches:

```ts
const handleSocketPinUpdate = useCallback((event: ChatWsEvent) => {
  if (event.chat_id !== chat.id) return;
  void refreshPins(false);
  setPinRefreshKey((previous) => previous + 1);
}, [chat.id, refreshPins]);
```

`frontend/src/components/chat/ChatWindow.tsx`. The event is a cache-invalidation signal; the server stays the only source of truth. Out-of-order events therefore cannot produce "an unpin arriving before the pin" — both orderings end in a refetch, and the refetch returns the committed state. A `pin_version` in the payload would guard a path that never reads the payload, and it would need a new column on `PinnedMessage`, which currently has none.

Handler idempotency follows from the same design: N events produce N refetches and one final state.

## The ordering bug was real, one layer down — fixed

You were right that there is a race here; it is in the refetch rather than the event stream. `refreshPins` and the drawer's `fetchPins` had no request-sequence guard, so:

- a rapid pin/unpin sequence puts several `listPins` calls in flight and an older response resolving last overwrites a newer list;
- switching channels does not cancel the previous channel's in-flight request, so a late reply renders one channel's pins inside another.

The drawer's copy was the more exposed of the two, because its effect re-runs on `pinRefreshKey`, which is bumped by every incoming `pin_update` — so any member's pin activity, not just the local user's, could trigger overlapping fetches.

Both now use a generation counter; only the newest request may write. The same sweep found and fixed the drawer's `listChatFiles` and `listScheduledMessages` loaders, which had the same cross-channel problem.

## Metrics — done

`chat_broadcast_publish_failures_total{event="pin"}` is new. Queueing successfully said nothing about the publish that follows it, so a worker that could not reach the channel layer failed after the request had already returned 201. The counter now increments per unreachable recipient and once for a whole-task failure, alongside the existing `chat_broadcast_enqueue_failures_total` for the enqueue side.

## Retries — deliberately not added

`pin_update` is a cache-invalidation signal, and the client refetches pins on every socket open. A dropped event costs a member the live update until their next reconnect or channel switch; it does not corrupt state. Retrying the whole broadcast would re-send to the recipients that already succeeded, and since every recipient reacts by refetching, the retry buys nothing the next reconnect would not. I would rather leave the failure visible in the counter above than add a retry path that is almost never the difference between correct and incorrect.

If you disagree I will add it — it is a small change — but I did not want to add a retry loop on a best-effort signal without saying why.

## Tests — added

Beyond the two commit-hook tests already present:

- `test_queued_pin_broadcast_excludes_the_member_who_made_the_change`
- `test_removing_a_pinned_message_broadcasts_the_unpin` (parametrized over delete and revoke)
- `test_removing_an_unpinned_message_broadcasts_nothing`

That middle one closes a gap your comment led me to: deleting or revoking a message dropped its pin row **without** broadcasting, so every other member's banner kept offering a jump to a message that no longer existed until they reloaded. The PR established live propagation for pin and unpin but left the third path that mutates pins silent. Both paths now go through `queue_pin_update`, and only when a pin row was actually removed.

While there, I also found that `notify_pin_update` was broadcasting back to the member who made the change (unlike `notify_reaction_update`, which excludes the actor), so a manager's own pin briefly marked their own channel as having an unseen pin. It now takes `actor_user_id` and excludes them; the argument defaults to `None` so tasks queued by the previous release still run during a deploy.

---

# 2. Migration and manager fallback

## Destructive migration — concern accepted, lighter implementation

The migration now logs every row before deleting it, with a summary line:

```
0013_remove_non_group_pinned_messages: deleting pin id=… chat=… message=… pinned_by=… created_at=…
0013_remove_non_group_pinned_messages: deleted N non-group pin(s) across M chat(s)
```

That gives the audit trail and the hand-recovery path your comment asks for, at five lines and no schema change.

I did not go with the archive table, for three reasons — happy to be overruled:

1. **Multi-tenant.** Chat tables live per tenant schema. An archive table is created in every tenant schema, and the "admin report listing affected chat IDs" has to aggregate across all of them, so it is not the single query the SQL sketch implies.
2. **The second step needs an owner.** "Hard-delete in a follow-up after review" is a management command plus a scheduled operational task plus someone who remembers. Realistically the archive table outlives the review and becomes a table nobody is willing to drop — trading a definite piece of technical debt for recoverability we can get from the log plus a backup.
3. **Scale.** On the dev database the whole table holds 4 pins and 0 of them are non-group. I cannot see preview/prod from here, so this is not proof — but it does suggest we are protecting a handful of rows.

If you still want the archive table, say so and I will add it, but I would want the follow-up deletion to have a named owner and a date rather than "after verification".

## Manager fallback — this PR does not introduce it

`ChatService.is_channel_manager` and its `get_fallback_manager_user_id` helper exist verbatim on `prod-preview` (`backend/chat/services.py`). `git diff origin/prod-preview...HEAD -- backend/chat/services.py` shows no change to either.

What the PR changes is the **frontend**, which previously had three divergent inline manager checks while the backend was already enforcing the three-level rule. So the fallback was already granting those permissions; the UI just disagreed with the server about who had them. `isChannelManager` mirrors the backend rule so the two stop diverging. That removes an inconsistency rather than granting anything new.

On tightening it — the numbers argue against doing it here. From the dev database:

| | |
|---|---:|
| group channels | 14 |
| with an explicit manager | 3 |
| **no explicit manager** | **11** |
| of those, with a `created_by` | 1 |
| of those, with no creator either | 10 |

Roughly four out of five group channels currently depend on the fallback, and ten of them have no creator to fall back to before it. If the fallback is removed, those channels have **no administrator at all** — nobody can rename them, edit the topic, pin, or add a manager. And promoting a manager is itself a manager-only action (`views.py`, "Only channel managers can assign managers"), so there is no in-app way out: it would need a data migration that appoints a manager for every ownerless channel.

That is a change to the permission model of every existing channel, and it belongs in its own ticket with that data migration attached — not inside an 88-file PR whose subject is pinned messages and delivery reliability. I am happy to open it and carry the numbers over; I would want the same query run against preview and production first, since the dev sample is small.

## The tests you asked for

Both already exist:

- non-group pins unreachable — `test_list_pins_excludes_direct_message_chats` (the list endpoint returns `[]` for a DM) and `test_private_chat_participants_cannot_pin_or_unpin` (403 on write).
- channels with no explicit manager — `test_legacy_fallback_manager_can_pin` pins the current behaviour, so if we do change the fallback later, that test is what will fail and force the decision to be explicit.

---

## Verification

- `pytest chat/tests` — **214 passed**
- `pytest chat/tests/test_pinned_messages.py` — **32 passed**
- `tsc --noEmit` — pre-existing errors in unrelated `e2e/` specs only; **0 in changed files**
- Jest — **858 passed**

quanwangniuniu (changes requested):
approved

quanwangniuniu (approved):
good job

PR #779, opened 2026-08-12
fix(agent): redact API keys from debug log paths

Joestar577:
## Summary

Fixes MED-334 by removing API key leakage from agent debug log paths. `llm_client.py` and `gemini_client.py` embed the Gemini API key in the request URL as `?key={api_key}`, which is required by the Gemini API, but `requests`/`urllib3` can log full URLs and headers at DEBUG level, and raw exception text (e.g. `str(exc)`) can carry that URL into tracebacks and DB-persisted error logs. This PR adds a central redaction layer rather than removing the key from the URL (not possible — Gemini requires it) or suppressing the logs entirely (loses debuggability).

The approach is a `logging.Filter` (`agent/log_redaction.py`) wired into Django's `LOGGING` config so it applies to every handler, including third-party loggers like `urllib3`, plus two targeted call-site fixes where a key could otherwise leak into a raised exception message or a database row.

## Backend

- Added `agent/log_redaction.py` — `RedactSecretsFilter(logging.Filter)` with two-layer protection: Layer 1 walks `record.args` (dicts/lists/tuples, including the dict-collapsed case Python's logging module uses for `logger.debug("...%s", {...})`) and masks values under sensitive header keys (`authorization`, `x-api-key`, `cookie`, etc., case-insensitive); Layer 2 regex-scrubs `record.msg` for `Bearer <token>`, `sk-*`, `AIzaSy*`, and `?key=`/`?token=` query params. The filter always returns `True` — it sanitises, never suppresses. Also exports `redact_headers()` and `redact_string()` as standalone helpers for direct use.
- Wired `redact_secrets` into `backend/settings.py`'s `LOGGING['filters']`, attached to the console handler and the root logger, so DEBUG-level output from `requests`/`urllib3` is covered, not just our own `logger.*` calls.
- `agent/gemini_client.py` — kept the `?key={api_key}` URL construction as-is (required by the Gemini API), but `_gemini_request_with_retry`'s `RuntimeError` on HTTP/network failure is now built with `redact_string()`. This matters because `requests.exceptions.ConnectionError`'s own `str()` embeds the full failed URL (including the key) — verified this leaks in the raw exception and is scrubbed in what we re-raise.
- `agent/llm_client.py` — `call_llm()`'s failure path (line 135) now writes `error_message=redact_string(str(exc))[:500]` instead of the raw `str(exc)[:500]`, so a leaked key can no longer land in `LLMCallLog.error_message`.

## Frontend

No frontend changes.

## Tests

    - `docker compose -f docker-compose.dev.yml --env-file .env exec backend pytest agent/tests/test_log_redaction.py --no-cov -v`
    - Result: 8 passed
    - (`--no-cov` used to avoid tripping the global coverage threshold gate on a targeted run)

- Django backend regression pytest:
    - `docker compose -f docker-compose.dev.yml --env-file .env exec backend pytest agent/test_call_llm.py agent/tests/ --no-cov -q`
    - Result: 25 passed (confirms no behavior change to quota reservation/commit/logging flow in `call_llm`)

- Django config check:
    - `docker compose -f docker-compose.dev.yml --env-file .env exec backend python manage.py check`
    - Result: System check identified no issues

- Frontend targeted unit tests: N/A — no frontend files changed.

- Diff hygiene:
    - `git diff --check`
    - Result: clean

## QA

- Demo video and QA checklist submitted in Discord `#qa-feedback`.

## Notes

- Migrations: no.
- No Vault integration — explicitly out of scope per the ticket.
- The `?key={api_key}` query param itself is intentionally left in the request URLs in both clients (required by the Gemini API); only its leakage into logs/errors is fixed.
- Manually verified end-to-end with simulated `requests.exceptions.ConnectionError` (the kind urllib3 raises on a real network failure, embedding the full URL+key) flowing through both `gemini_client._gemini_request_with_retry` and `llm_client.call_llm` → `LLMCallLog.error_message`; confirmed the key is redacted at both points.

### DIFF
--- a/backend/agent/gemini_client.py
+++ b/backend/agent/gemini_client.py
@@ -12,6 +12,8 @@ import time
 import requests
 from django.conf import settings
 
+from agent.log_redaction import redact_string
+
 logger = logging.getLogger(__name__)
 
 GEMINI_MODEL = "gemini-2.5-flash-lite"
@@ -69,10 +71,10 @@ def _gemini_request_with_retry(
             if status_code == 429:
                 raise GeminiRetriesExhausted("Gemini rate limited (HTTP 429).") from exc
             raise RuntimeError(
-                f"Gemini request failed with HTTP {status_code or 'unknown'}."
+                redact_string(f"Gemini request failed with HTTP {status_code or 'unknown'}.")
             ) from exc
         except requests.exceptions.RequestException as exc:
-            raise RuntimeError("Gemini network error.") from exc
+            raise RuntimeError(redact_string(f"Gemini network error: {exc}")) from exc
     raise GeminiRetriesExhausted("Gemini rate limited (HTTP 429).") from last_http_error
 
 
--- a/backend/agent/llm_client.py
+++ b/backend/agent/llm_client.py
@@ -11,6 +11,7 @@ from typing import Any
 
 from django.conf import settings
 
+from agent.log_redaction import redact_string
 from stripe_meta.exceptions import QuotaError
 from stripe_meta.models import LLMCallLog
 from stripe_meta.services import (
@@ -132,7 +133,7 @@ def call_llm(
             output_cost_cents=0,
             total_cost_cents=0,
             success=False,
-            error_message=str(exc)[:500],
+            error_message=redact_string(str(exc))[:500],
         )
         raise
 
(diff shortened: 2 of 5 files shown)


quanwangniuniu (approved):
Good job

PR #783, opened 2026-08-15
feat(observability): Grafana dashboard for WebSocket session counts — MED-319

zizhenliu0427:
## Summary

**MED-319 — Grafana dashboard for WebSocket session counts**

[MED-319 — Chat: Grafana dashboard for WebSocket session counts](https://linear.app/mediajira/issue/MED-319/chat-grafana-dashboard-for-websocket-session-counts)

Adds a `ws_sessions_active{channel="..."}` gauge, exported through the existing `django_prometheus` endpoint, and a Grafana dashboard built on it. A new Channels consumer base class owns the counting, so every WebSocket consumer reports its live session count without touching its own `connect`/`disconnect` logic.

Prometheus and Grafana already exist in `docker-compose.dev.yml` and `devops/grafana/` already holds three dashboards, so this PR adds a metric and a fourth board rather than standing up monitoring.

The platform was entirely pre-existing — `django_prometheus`, the `/metrics` endpoint, the `django` scrape job and three dashboards have been in place since December 2025 — but nothing this ticket asks for was. `ws_sessions_active` appears nowhere in the history before this branch, `backend/core/metrics.py` does not exist on `prod-preview`, and no consumer base class existed: all seven consumers subclassed `AsyncWebsocketConsumer` directly. Worth setting expectations on where the effort actually went: defining the metric is about seven lines and the board is largely template work. The substance was making the count correct on the paths the ticket does not mention — rejected handshakes, post-accept crashes, repeated closes and sustained churn.

### Acceptance criteria against production reality

"Metric exported per consumer" is satisfied for **all six consumers that are reachable in production**. The fourth consumer named in the ticket's Context, `retrospective`, has no reachable endpoint at all: it is not registered in `backend/asgi.py`, so the module is never imported and no series is ever created. That is a pre-existing routing and security issue, detailed under Follow-up, and is unrelated to the dashboard work.

It is worth noting that the ticket's list of four does not match the production topology. Three consumers that genuinely run — `spreadsheet`, `csm` and `portal` — are absent from it, and the one that does not run is present. The list looks like it was drawn from which apps have a `consumers.py` file rather than from which endpoints are actually routed.

### The counting design

The whole value of this ticket rests on the gauge being trustworthy — it will be used for capacity decisions and, in a follow-up ticket, for alert thresholds. A gauge that drifts is worse than no gauge, because people act on it. Two failure modes were found and closed by construction:

- **Counting in `connect()` drives the gauge negative.** Consumers reject handshakes by calling `close()` and returning *before* `accept()` (auth failure `4001`, authorisation failure `4003`). Channels still delivers `websocket.disconnect` afterwards, so a naive increment/decrement pair decrements a session that was never counted. Reverting to that implementation reproduces `-1.0` on a single rejected handshake.
- **Releasing in `disconnect()` leaks the session permanently.** `disconnect()` only runs on an orderly shutdown; a consumer that raises after accepting never reaches it. `chat` performs post-accept initialisation and `spreadsheet` runs an authorisation watchdog, so this path is real. Reverting to that implementation leaks one session per crash and accumulated to `3.0` across test retries.

The session is therefore counted in `accept()` — after the handshake actually succeeds — and released in `__call__`'s `finally`, which wraps the entire connection and runs on every exit path including exceptions and cancellation. An idempotent flag makes both sides exactly-once, so repeated `accept()` or repeated `close()` cannot double count. `chat` closes in six places and `spreadsheet` in seven, so that guard is load-bearing rather than defensive padding.

### Double counting and data integrity

Raised in standup as the main risk on this ticket, so it was audited specifically rather than assumed. Note that the leak above *is* a double count in effect: a dead session still occupying the gauge while the user reconnects makes one person read as two.

Every way a session could be counted or released more than once, and what closes it:

| Path | Guarantee | Evidence |
|---|---|---|
| `accept()` called twice | Flag makes the increment exactly-once | `test_accepting_twice_counts_one_session` |
| `close()` called twice | Flag makes the release exactly-once; a second release would read `-1`, not `0` | `test_closing_twice_releases_one_session` |
| Rejected handshake still receives `websocket.disconnect` | Never counted, so never released | `test_rejected_handshake_is_never_counted`, reproduced at `-1.0` without the guard |
| Consumer raises after `accept()` | `finally` releases on every exit path | `test_session_is_released_when_the_consumer_crashes_after_accepting`, reproduced at `3.0` without it |
| Sustained churn accumulating fractions | 25 cycles return to exactly the baseline | `test_repeated_connect_disconnect_cycles_leave_no_drift` |
| A consumer bypassing the base class | No consumer overrides `accept`, `__call__`, `websocket_connect` or `websocket_disconnect`; each calls `accept()` exactly once | Verified by grep across all seven |
| The same process scraped by two jobs | `prometheus.yml` has one `django` target; the `nextjs` job scrapes a different application | Verified in the scrape config |

A state gauge helps here in a way an event counter cannot: because it must return to zero when everyone disconnects, a double count cannot hide inside a growing total — it shows up as a number that will not settle. That is what the churn test and the real-server probe below exercise.

#### Conclusion on integrity as a whole

Integrity here means one thing: **the number on the board equals the number of live sessions in the process.** Taking the threats to that one at a time rather than only the double-count case:

| Threat | Status |
|---|---|
| Over-count — repeated `accept()` or `close()` | Closed by the idempotent flag; two tests |
| Under-count — gauge driven negative by a rejected handshake | Closed by counting only after `accept()`; reproduced at `-1.0` first |
| Leak — counted but never released | Closed by releasing in `__call__`'s `finally`; reproduced at `3.0` first |
| Drift — fractional accumulation under churn | 25-cycle test plus 30 real cycles, exact return to baseline |
| Bypass — a consumer not going through the base class | No consumer overrides `accept`/`__call__`/`websocket_connect`/`websocket_disconnect`; enforced by a test |
| Missing series — a channel absent rather than zero | Registered at import by `__init_subclass__`; a consumer that omits `ws_channel` fails at import |
| Unrouted consumer counted as covered | Coverage read from the real ASGI router, not from test imports |
| Scrape-layer duplication | One `django` target in `prometheus.yml`; the `nextjs` job is a different application |

**Within a backend process, the gauge is exact.** Every failure mode above is closed by construction and pinned by a test, and the two most important were reproduced against the wrong implementation before being fixed, so the tests are known to catch them rather than merely passing.

Two residual risks are stated rather than claimed, because a single local daphne process cannot close them:

- **Cross-instance duplication during a rolling deployment.** Panels sum across instances, which is correct while instance labels are stable. A deployment that changes them could double count for the length of the staleness window. Not reproducible in dev.
- **A leak rate below the probe's resolution.** 65 handshakes over a few minutes would not reveal a leak of one connection in ten thousand. That is only observable by watching the graph over days — which is what this ticket exists to make possible.

And one case that looks like a defect but is not: a client that reconnects before the server has noticed the dead socket is counted twice until the old connection times out. Inherent to any connection gauge, self-clearing, and stated in the dashboard description.

### Scope note — seven consumers, not four

The ticket's Context names chat, asset, meetings and retrospective. The base class covers all seven WebSocket consumers, adding csm, portal and spreadsheet.

This is the cheaper reading of the requested work rather than an expansion of it. The Scope line asks for a *consumer base class*, and a base class applies to whoever inherits it; the acceptance criterion says "Metric exported per consumer". Restricting to four would mean deliberately leaving three consumers on the old base class for no technical reason, which makes the base class non-universal and silently denies metrics to any consumer added later. The marginal cost was one line each and no extra dashboard work, because the board aggregates by label. The three additions are also the highest-value ones operationally: spreadsheet holds the heaviest connections (presence, cursors, watchdog) and csm/portal are the customer-facing paths.

## Files Changed

**Backend**

- `backend/core/metrics.py` — new. The `ws_sessions_active` gauge, labelled by `channel`
- `backend/core/consumers.py` — new. `InstrumentedAsyncWebsocketConsumer`; counts in `accept()` (L53-57), releases in `__call__` (L45-51), registers the label series at import time via `__init_subclass__` (L34-43) and raises `TypeError` if a subclass omits `ws_channel`
- `backend/{asset,chat,csm,meetings,portal,retrospective,spreadsheet}/consumers.py` — inherit the base class and declare `ws_channel`; one import line and one attribute each, no change to any `connect`/`disconnect` body
- `backend/core/tests/test_ws_session_metrics.py` — new suite, 10 tests. Coverage of the routed set is taken from the real `backend.asgi` URL router rather than by importing the consumers: importing one registers its series in the test process, so a suite that imports them directly creates the series it claims to verify and reports coverage production does not have

**Dashboard**

- `devops/grafana/websocket-sessions-dashboard.json` — new. Six panels: current total, per-channel bar gauge, peak over the selected range, per-channel time series, per-instance time series, and a channel/instance table

## Expected Result

- `/metrics` exposes one `ws_sessions_active{channel="..."}` series per routed consumer, reporting `0` while idle rather than being absent
- The gauge rises on an accepted handshake and returns to baseline on disconnect
- A rejected handshake is never counted and cannot drive the gauge negative
- A consumer that crashes after accepting still releases its session
- Sustained connect/disconnect churn leaves no drift
- The dashboard shows live counts split by channel and by backend instance

## Test

- [x] `core/tests/test_ws_session_metrics.py` — **10 passed**
- [x] Existing WebSocket suites (`asset`, `chat`, `spreadsheet` ×2, `retrospective`) — **no new failures**, verified by baseline comparison below
- [x] Live exposition check — `/metrics` on the running dev backend lists six channel series
- [x] Prometheus scrape — the `django` target is `up` and the series are queryable
- [x] Grafana — dashboard imports and renders against the Prometheus datasource
- [x] PromQL — all six panel expressions pass `promtool check rules`
- [ ] PR CI on `prod-preview`

### Baseline comparison for the pre-existing failures

The WebSocket suites report 2 failures and 36 errors in the local environment. These were confirmed pre-existing by stashing the seven consumer changes and running the identical command on the clean tree:

| Run | passed | failed | errors |
|---|---|---|---|
| Clean `prod-preview` | 62 | 2 | 36 |
| With this branch | 68 | 2 | 36 |

The failure and error counts are identical; the six extra passes are this branch's new tests. The errors are tenant-schema `_fixture_teardown` flush failures and test-database contention (pgbouncer holds pooled connections to `test_mediajira_db`, so the next run cannot drop it), neither of which this branch touches. Running with `DB_HOST=db` to bypass pgbouncer does not change the counts.

### The two regression tests were verified to have teeth

Both counting bugs described above were reproduced deliberately before being fixed, so the tests are known to fail against the wrong implementation rather than passing incidentally:

- Restoring the `connect()`/`disconnect()` counting pair makes `test_rejected_handshake_is_never_counted` fail with `assert -1.0 == 0`
- Restoring the `websocket_disconnect()` release makes `test_session_is_released_when_the_consumer_crashes_after_accepting` fail with `assert 3.0 == 0`

### Integrity check against real daphne

The unit tests use `WebsocketCommunicator`, which does not exercise the real protocol stack. A probe issuing raw WebSocket handshakes against the running daphne process was run separately — 65 handshakes in total, zero drift:

| Scenario | Result |
|---|---|
| 20 concurrent sessions | `20.0`, returns to `0.0` when closed |
| 30 rapid connect/disconnect cycles | `0.0` — no drift |
| 15 rejected handshakes (bad token) | `0.0` — rejections do not corrupt the count |
| Rejected and accepted interleaved, 10 each | `10.0` held, `0.0` after release |

## Out of Scope / Follow-up

- **Alert rules** — excluded by the ticket, and correctly so: no sensible threshold can be chosen before a baseline exists. This ticket produces the baseline; alerting should follow after the graph has been observed.
- **`RetrospectiveConsumer` is unrouted dead code with an authentication hole, and routing it would gain nothing.** Investigated rather than assumed, because it is the one consumer named in the ticket that produces no data:

  - **Not routed.** `retrospective/routing.py` exists but `backend/asgi.py` never imports it, so the module is never loaded and no series is created. Only six channels appear in the live exposition.
  - **Nothing connects to it.** There is no `ws/retrospective` reference anywhere in the frontend, and no retrospective page directory at all.
  - **Nothing broadcasts to it.** The only code that sends to a `retrospective_{id}` group is the consumer's own `receive()`. No service, task or signal anywhere in the backend publishes to it.
  - **Authentication is disabled.** `connect()` never rejects an unauthenticated user: the checks are commented out with "For testing, skip authentication", and the surviving branch only fires when `scope['user']` is `None`, which the middleware never produces — it sets `AnonymousUser`. Execution falls through to `accept()`. Every other consumer closes with `4001`/`4003` there. The existence check is commented out too, so any id is accepted.
  - **`receive()` accepts client-initiated broadcasts.** Any connected client can send `group_broadcast`, `retrospective_completed`, `insight_generated` or `kpi_update` and have it relayed to the entire group. Combined with the missing authentication, an unauthenticated caller who guesses an id could fabricate completion and KPI events for every connected client.

  So this is not a dashboard omission. Routing it as it stands would publish an unauthenticated injection endpoint, and routing it *after* fixing the authentication would still report a permanent `0`, because nothing connects to it and nothing broadcasts to it. The honest question is whether the consumer should exist at all, which belongs in its own ticket — most likely as a deletion rather than a repair.

  **The metric side is nevertheless finished for it, so whoever fixes the consumer does not have to come back through this work.** `RetrospectiveConsumer` already inherits the instrumented base and declares its channel, and the dashboard discovers channels through `label_values(ws_sessions_active, channel)` with no channel name hardcoded in any of the six panels. Registering the route is the only change required.

  This was verified rather than assumed. Adding the route to `backend/asgi.py` as a temporary experiment — with no change to the metric or the dashboard — produced the series immediately:

  ```
  ws_sessions_active{channel="retrospective"} 0.0
  ```

  and the suite failed with the instruction the next developer needs, before the experiment was reverted:

  ```
  AssertionError: missing from backend/asgi.py: [];
  now routed, remove from KNOWN_UNROUTED_CHANNELS: ['retrospective']
  ```

  That is the intended handover: `test_no_consumer_is_silently_left_out_of_the_asgi_router` records the exemption so a *newly* unrouted consumer fails instead of passing unnoticed, and it announces itself the moment retrospective becomes routed, so the exemption cannot outlive the fix. `test_unrouted_consumers_are_already_wired_for_when_they_are_routed` holds the other half — an unrouted consumer must still be instrumented, so routing stays a one-line fix rather than a metrics task as well.
- **No monitoring stack outside dev.** `grafana` and `prometheus` appear only in `docker-compose.dev.yml`; all production and preview compose files have zero references, and there is no deployment configuration for them elsewhere in the repo. The dashboard JSON is committed as the ticket requires, but deploying Prometheus/Grafana against preview or production and importing the board needs its own ticket — otherwise the metric is collected nowhere and the board has no host.
- **Dashboards are not provisioned.** The grafana service mounts no volumes, so a manually imported board is lost when the container is recreated. Restoring the `provisioning/` directory removed by `8927eee` (`revert(infra): defer Grafana monitoring work to MED-319`) would fix this. That commit also removed a chat monitoring dashboard and alert rules; those cover outbox backlog, Celery queue depth and Postgres connections, which is a different concern from session counts, and alert rules are explicitly out of scope here. Worth confirming whether restoring them belongs to this ticket or a separate one.
- **The metric counts sessions, not users.** One person with three tabs is three sessions, and the same person in chat and spreadsheet appears once per channel. A client that reconnects before the server notices the dead socket is briefly counted twice until the old connection times out. Both are documented in the dashboard description so the numbers are not misread as a user count.
- **Multi-instance behaviour is untested locally.** Only one daphne process exists in dev. Panels sum across instances, which is correct while instance labels are stable; a rolling deployment that changes them could double count for the length of the staleness window.

Ticket: [MED-319](https://linear.app/mediajira/issue/MED-319/chat-grafana-dashboard-for-websocket-session-counts)

### DIFF
--- a/backend/asset/consumers.py
+++ b/backend/asset/consumers.py
@@ -1,16 +1,18 @@
 import json
-from channels.generic.websocket import AsyncWebsocketConsumer
 from channels.db import database_sync_to_async
 from django.contrib.auth.models import AnonymousUser
+from core.consumers import InstrumentedAsyncWebsocketConsumer
 from .models import Asset
 
 
-class AssetConsumer(AsyncWebsocketConsumer):
+class AssetConsumer(InstrumentedAsyncWebsocketConsumer):
     """
     WebSocket consumer for real-time asset updates.
     Handles connections to asset-specific groups for live updates.
     """
-    
+
+    ws_channel = 'asset'
+
     async def connect(self):
         """Handle WebSocket connection"""
         self.asset_id = self.scope['url_route']['kwargs']['asset_id']
--- a/backend/chat/consumers.py
+++ b/backend/chat/consumers.py
@@ -7,7 +7,6 @@ from collections import OrderedDict
 from contextlib import suppress
 from datetime import timedelta
 from functools import wraps
-from channels.generic.websocket import AsyncWebsocketConsumer
 from channels.db import database_sync_to_async
 from asgiref.sync import sync_to_async
 from django.contrib.auth import get_user_model
@@ -16,6 +15,7 @@ from django.core.cache import cache
 from .models import Chat, ChatParticipant, Message, MessageStatus
 from .realtime import broadcast_event_to_user_groups
 from django.conf import settings
+from core.consumers import InstrumentedAsyncWebsocketConsumer
 from core.tenant_context import tenant_schema_context, validate_tenant_schema
 from .services import (
     ChatService,
@@ -65,7 +65,7 @@ def tenant_db_method(method):
     return wrapped
 
 
-class ChatConsumer(AsyncWebsocketConsumer):
+class ChatConsumer(InstrumentedAsyncWebsocketConsumer):
     """
     WebSocket consumer for real-time chat functionality.
     
@@ -85,7 +85,9 @@ class ChatConsumer(AsyncWebsocketConsumer):
     - typing_indicator: Someone is typing
     - error: Error occurred
     """
-    
+
+    ws_channel = 'chat'
+
     async def connect(self):
         """Handle WebSocket connection"""
         self.user_id = self.scope['url_route']['kwargs']['user_id']
--- a/backend/core/consumers.py
+++ b/backend/core/consumers.py
@@ -1,0 +1,57 @@
+"""Base Channels consumer that reports live session counts to Prometheus."""
+
+from channels.generic.websocket import AsyncWebsocketConsumer
+
+from .metrics import ws_sessions_active
+
+
+class InstrumentedAsyncWebsocketConsumer(AsyncWebsocketConsumer):
+    """AsyncWebsocketConsumer that keeps ``ws_sessions_active`` accurate.
+
+    Subclasses only declare ``ws_channel``; their ``connect``/``disconnect``
+    stay untouched. The counting deliberately hangs off ``accept()`` and
+    ``__call__`` rather than the ``connect``/``disconnect`` hooks:
+
+    - ``connect()`` is the wrong place to count. Consumers reject handshakes by
+      calling ``close()`` and returning before ``accept()``, and Channels still
+      delivers ``websocket.disconnect`` afterwards, so counting every connect
+      would let rejected handshakes decrement the gauge below zero.
+    - ``disconnect()`` is the wrong place to release. It only runs on an orderly
+      shutdown: a consumer that raises after accepting (or is cancelled) never
+      reaches it, and would stay counted until the process restarted. ``__call__``
+      wraps the whole connection, so every exit path releases the session.
+
+    A session is therefore counted once the handshake is actually accepted, and
+    released exactly once when the connection ends, however it ends.
+    """
+
+    #: Value of the ``channel`` label, e.g. ``"chat"``. Required.
+    ws_channel = None
+
+    # Class-level default: subclasses need no __init__ to get a starting value.
+    _ws_session_counted = False
+
+    def __init_subclass__(cls, **kwargs):
+        super().__init_subclass__(**kwargs)
+        if not cls.ws_channel:
+            raise TypeError(
+                f'{cls.__name__} must set ws_channel to label its '
+                f'ws_sessions_active series'
+            )
+        # Instantiate the child gauge at import time so an idle channel reports
+        # 0 instead of dropping out of /metrics until its first connection.
+        ws_sessions_active.labels(channel=cls.ws_channel)
+
+    async def __call__(self, scope, receive, send):
+        try:
+            await super().__call__(scope, receive, send)
+        finally:
+            if self._ws_session_counted:
+                self._ws_session_counted = False
+                ws_sessions_active.labels(channel=self.ws_channel).dec()
+
+    async def accept(self, *args, **kwargs):
+        await super().accept(*args, **kwargs)
+        if not self._ws_session_counted:
+            self._ws_session_counted = True
+            ws_sessions_active.labels(channel=self.ws_channel).inc()
--- a/backend/core/metrics.py
+++ b/backend/core/metrics.py
@@ -10,8 +10,22 @@ from prometheus_client import Gauge
 
 # Sessions that completed the WebSocket handshake and have not disconnected
 # yet. Rejected handshakes never count: connect() closes those before accept().
+#
+# Cardinality: ``channel`` is bounded by the number of consumer classes, not by
+# traffic. Its only source is the class-level ``ws_channel`` constant that each
+# consumer declares (see InstrumentedAsyncWebsocketConsumer); the label is never
+# derived from the connection -- not from scope, url_route kwargs, the user, the
+# room or the sheet id -- so no volume of sessions can mint a new series. The
+# allowed values are exactly:
+#
+#     asset | chat | csm | meetings | portal | retrospective | spreadsheet
+#
+# Adding a per-room or per-user label would mean editing the base class, and
+# test_channel_label_set_is_bounded_and_declared fails on any label value that
+# is not in that list. Keep it that way: a label whose values track connections
+# is what turns a gauge into a series explosion.
 ws_sessions_active = Gauge(
     'ws_sessions_active',
     'Currently connected WebSocket sessions',
-    ['channel'],  # channel: chat | spreadsheet | csm | portal | meetings | retrospective | asset
+    ['channel'],
 )
--- a/backend/core/metrics.py
+++ b/backend/core/metrics.py
@@ -1,0 +1,17 @@
+"""Prometheus metrics shared across apps.
+
+Scraped through django_prometheus alongside the rest of the application's
+metrics. The backend serves HTTP and WebSocket traffic from a single daphne
+process (see backend/Dockerfile), so an in-process gauge and the /metrics
+endpoint always agree; with several replicas, sum across instances in Grafana.
+"""
+
+from prometheus_client import Gauge
+
+# Sessions that completed the WebSocket handshake and have not disconnected
+# yet. Rejected handshakes never count: connect() closes those before accept().
+ws_sessions_active = Gauge(
+    'ws_sessions_active',
+    'Currently connected WebSocket sessions',
+    ['channel'],  # channel: chat | spreadsheet | csm | portal | meetings | retrospective | asset
+)
--- a/backend/core/tests/test_ws_session_metrics.py
+++ b/backend/core/tests/test_ws_session_metrics.py
@@ -1,0 +1,134 @@
+"""Tests for the ws_sessions_active gauge on the instrumented consumer base."""
+
+import pytest
+from channels.testing import WebsocketCommunicator
+from prometheus_client import REGISTRY
+
+from asset.consumers import AssetConsumer
+from chat.consumers import ChatConsumer
+from core.consumers import InstrumentedAsyncWebsocketConsumer
+from csm.consumers import CsmConversationConsumer
+from meetings.consumers import MeetingDocumentConsumer
+from portal.consumers import PortalConversationConsumer
+from retrospective.consumers import RetrospectiveConsumer
+from spreadsheet.consumers import SheetConsumer
+
+WEBSOCKET_CONSUMERS = [
+    AssetConsumer,
+    ChatConsumer,
+    CsmConversationConsumer,
+    MeetingDocumentConsumer,
+    PortalConversationConsumer,
+    RetrospectiveConsumer,
+    SheetConsumer,
+]
+
+
+class AcceptingConsumer(InstrumentedAsyncWebsocketConsumer):
+    ws_channel = 'test_accepting'
+
+    async def connect(self):
+        await self.accept()
+
+
+class RejectingConsumer(InstrumentedAsyncWebsocketConsumer):
+    ws_channel = 'test_rejecting'
+
+    async def connect(self):
+        # Same shape as the real consumers' auth failures: close, no accept.
+        await self.close(code=4001)
+
+
+class CrashingConsumer(InstrumentedAsyncWebsocketConsumer):
+    ws_channel = 'test_crashing'
+
+    async def connect(self):
+        await self.accept()
+        raise RuntimeError('post-accept initialization blew up')
+
+
+@pytest.fixture(autouse=True)
+def in_memory_channel_layer(settings):
+    """Keep the consumers off the Redis layer these tests do not need."""
+    settings.CHANNEL_LAYERS = {
+        'default': {'BACKEND': 'channels.layers.InMemoryChannelLayer'}
+    }
+
+
+def active_sessions(channel):
+    return REGISTRY.get_sample_value('ws_sessions_active', {'channel': channel})
+
+
+async def test_accepted_connection_is_counted_until_disconnect():
+    communicator = WebsocketCommunicator(AcceptingConsumer.as_asgi(), '/ws/test/')
+
+    connected, _ = await communicator.connect()
+    assert connected
+    assert active_sessions('test_accepting') == 1
+
+    await communicator.disconnect()
+    assert active_sessions('test_accepting') == 0
+
+
+async def test_concurrent_sessions_are_counted_independently():
+    first = WebsocketCommunicator(AcceptingConsumer.as_asgi(), '/ws/test/')
+    second = WebsocketCommunicator(AcceptingConsumer.as_asgi(), '/ws/test/')
+    await first.connect()
+    await second.connect()
+    assert active_sessions('test_accepting') == 2
+
+    await first.disconnect()
+    assert active_sessions('test_accepting') == 1
+
+    await second.disconnect()
+    assert active_sessions('test_accepting') == 0
+
+
+async def test_rejected_handshake_is_never_counted():
+    """Channels delivers websocket.disconnect for rejected handshakes too.
+
+    Counting on connect() rather than accept() would decrement a session that
+    was never counted and drive the gauge negative.
+    """
+    communicator = WebsocketCommunicator(RejectingConsumer.as_asgi(), '/ws/test/')
+
+    connected, _ = await communicator.connect()
+    assert not connected
+    assert active_sessions('test_rejecting') == 0
+
+    await communicator.disconnect()
+    assert active_sessions('test_rejecting') == 0
+
+
+async def test_session_is_released_when_the_consumer_crashes_after_accepting():
+    """A consumer that raises post-accept never reaches disconnect().
+
+    Releasing there instead of around the whole connection would keep the
+    session counted until the process restarted.
+    """
+    communicator = WebsocketCommunicator(CrashingConsumer.as_asgi(), '/ws/test/')
+
+    await communicator.connect()
+    with pytest.raises(RuntimeError, match='post-accept'):
+        await communicator.wait(timeout=5)
+
+    assert active_sessions('test_crashing') == 0
+
+
+def test_every_websocket_consumer_exports_its_own_channel_series():
+    labels = [consumer.ws_channel for consumer in WEBSOCKET_CONSUMERS]
+    assert len(set(labels)) == len(labels), f'duplicate ws_channel labels: {labels}'
+
+    # Declaring a subclass registers the series, so a channel is scrapeable
+    # before its first connection instead of missing from /metrics. Asserting
+    # presence rather than 0: other test modules sharing this xdist worker may
+    # legitimately have sessions open.
+    for label in labels:
+        assert active_sessions(label) is not None
+
+
+def test_consumer_without_a_channel_label_is_rejected():
+    with pytest.raises(TypeError, match='ws_channel'):
+
+        class UnlabelledConsumer(InstrumentedAsyncWebsocketConsumer):
+            pass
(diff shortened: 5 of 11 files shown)


quanwangniuniu (changes requested):
Scope: backend/core/metrics.py and deployment (Prometheus / ASGI workers)
``` 
There may be a multi-process / multi-instance correctness gap: the current implementation appears to be a process-local Gauge updated inside the Channels consumer. In a multi-worker deployment (multiple Daphne/uvicorn processes, celery workers, or scaled replicas) an in-process Gauge will only reflect a single process' view and will not represent the global number of active WebSocket sessions.

Please either:
- Move the canonical session count to a central store (Redis) and expose a collector that reads those Redis keys and sets a Gauge for Prometheus; or
- Use a Prometheus approach that supports multi-process collection for your server (e.g. the prometheus_client multiprocess mode or a proper multiprocess collector configured for your ASGI server), and ensure Prometheus scrapes and aggregates across instances (use `sum by (channel)` in dashboards/alerts); or
- Replace the in-process gauge with an increment/decrement of a shared, atomic counter (Redis INCR/DECR) and expose/read that as the metric.

Also add an integration test or a local-compose test demonstrating correct aggregation across two simulated worker processes.
```
Why this matters:
A process-local metric misleads capacity planning and alerts in scaled environments. Prometheus scraping a single process or relying on non-aggregated gauges will undercount (or produce confusing per-process series). Fixing this prevents wrong capacity decisions and alert noise.

Scope: backend/asgi.py
``` 
The PR notes the `retrospective` consumer is not registered (module never imported), which indicates a routing / registration gap and an observability blind spot.

Please:
- Add the missing route for the `retrospective` consumer in backend/asgi.py (or the routing module you use), or explicitly import the module during startup if that consumer is intentionally not routed from the usual place.
- Add a small unit/integration test that asserts all expected consumer names/endpoints are present in the ASGI routing table (or that importing `backend.consners` triggers metric creation).
- If that endpoint is intentionally disabled, update the PR description and Grafana dashboard notes to mark it as unreachable and create a follow-up issue for routing/security remediation.

Include the route and test change in this PR or link the follow-up issue.
```
Why this matters:
If a consumer isn't imported or routed, it will never register metrics and your dashboard will give a false sense of coverage. Also, unregistered consumers may point to a separate security/routing problem that needs triage.

Scope: Metric labeling and Prometheus config (devops/grafana/, prometheus relabeling)
``` 
The metric uses a `channel="..."` label; please confirm that the set of channel label values is small and bounded. If `channel` can be high-cardinality (user IDs, room IDs, dynamic strings), this will cause unbounded metric series growth.

Please:
- Ensure `channel` is limited to a finite set (e.g., enum of known consumer types) before adding as a label; do not use per-session or per-user identifiers as labels.
- If you need per-target counts (e.g., per-room), expose only aggregated metrics (e.g., `ws_sessions_by_room_bucket`) or use exemplars/trace linking instead of labels.
- Add a Prometheus relabeling rule (or Grafana dashboard note) to trim or map dynamic label values to safe categories (e.g., map unknowns to "other").
- Add automated tests that assert /metrics output contains only the expected channel label set in dev, or add CI linting to detect new label values.

If the label set is already bounded, add a comment explaining the allowed values and why they are safe.
```
Why this matters:
High-cardinality labels cause Prometheus memory explosions and slow queries. Confirming and constraining label values prevents serious production monitoring problems.

---

zizhenliu0427:
Thanks for the review — all three are the right things to check on a metric that will feed capacity decisions.

**Where I landed:** points 2 and 3 I've changed, including the tests you asked for. On point 1 I'd like to put some evidence forward rather than change it, because I think the proposed fix would cost us the property this PR was built around — but I'm happy to be overruled. Taking them in your order. Suite is **14 passed**.

---

## 1. Multi-process gauge — the concern is right in general; I don't think the stack has it

**The premise doesn't currently hold.** There's one process per container and no scaling:

- `backend/Dockerfile:35` — `exec daphne -b 0.0.0.0 -p 8000 ... backend.asgi:application`. Daphne doesn't pre-fork workers.
- No `replicas:` in any production or preview compose file.
- Celery workers don't serve WebSockets and aren't scraped — `prometheus.yml` has one `django` target; the `nextjs` job is a different application.

So today the in-process gauge and `/metrics` are the same process by construction.

**On `prometheus_client` multiprocess mode.** That exists for pre-fork servers like gunicorn, where several worker processes share one scrape endpoint and the exposition has to be reassembled from a shared directory. Daphne is single-process, so there's nothing to reassemble. Across multiple *containers* it doesn't apply either — each gets its own scrape target and `instance` label, which is the case Prometheus already handles by aggregating at query time.

**On the shared Redis counter — this is the one I'd argue against on correctness, not effort.** It reintroduces the exact failure this PR was built to eliminate. The current design releases a session in `__call__`'s `finally`, which runs on every exit path including exceptions and cancellation; that's what fixed the permanent leak I reproduced at `3.0`. But no `finally` survives SIGKILL, OOM, or a hard container stop. With an in-process gauge that doesn't matter — the process dies, its series leaves the scrape, and `sum` drops correctly. It's self-healing. With `INCR`/`DECR` in Redis, a process killed with sessions open leaves those decrements permanently unpaid, and nothing can reconcile them: Redis doesn't know which sessions belonged to the dead process. The number drifts up forever and the only remedy is a manual reset. That trades a bounded, self-correcting error for an unbounded, sticky one, on a gauge people are meant to trust for capacity decisions.

**Your second option is already implemented.** All six panels aggregate — `sum by (channel)`, `sum by (instance)`, `sum(...)` for the totals — and the dashboard description already states "The gauge is per backend process, so every panel sums across instances". The per-instance panel exists precisely so uneven lines across replicas are visible.

**On the test you asked for — added.** `test_two_processes_aggregate_to_the_global_total` spawns two real processes, each with its own registry, holding a different number of sessions, and asserts the parts sum to the whole:

```
process A view: 3.0
process B view: 5.0
sum across instances: 8.0
parent process view: 0.0
```

Different counts on purpose: a result that accidentally reflected a single process' view wouldn't equal the total. This is your stated concern reproduced directly — each instance does report only its own sessions — and it shows the aggregation the dashboard performs is what closes it.

**Where I think you're straightforwardly right,** and it's in the PR under Out of Scope: during a rolling deployment that changes instance labels, a stale and a fresh instance can both be counted for the length of the staleness window. That's real, it's not reproducible on one local daphne, and it wants a look once a stack is actually deployed — which is its own blocker, since `prometheus` and `grafana` currently exist only in `docker-compose.dev.yml` and nothing collects this metric outside a developer's machine.

If we do move to multiple replicas per environment, I'd revisit this — but I'd still expect `sum by (channel)` over per-instance gauges to be the answer, with the shared-counter approach reserved for something that genuinely needs a single global value.

## 2. Retrospective — agreed, taking your third option

Your option 3 is what I think is correct here, and I've now completed the parts I'd left implicit.

**Why not option 1 (add the route).** `RetrospectiveConsumer.connect()` never rejects an unauthenticated user — the checks are commented out with "For testing, skip authentication", and the surviving branch only fires when `scope['user']` is `None`, which the middleware never produces (it sets `AnonymousUser`). Execution falls through to `accept()`. Every other consumer closes with `4001`/`4003` there. Separately, `receive()` relays client-initiated `group_broadcast`, `retrospective_completed`, `insight_generated` and `kpi_update` to the whole group, so an unauthenticated caller who guesses an id could fabricate completion and KPI events for every connected client.

Routing it would turn a piece of dead code into a reachable unauthenticated injection endpoint. And it would still report a permanent `0`: nothing in the frontend connects to `ws/retrospective` (no reference, no page directory) and nothing in the backend broadcasts to a `retrospective_{id}` group except the consumer's own `receive()`. So it's a security regression in exchange for a flat line.

**On the test you asked for** — I think this one already exists, so let me point at it rather than add a second: `test_no_consumer_is_silently_left_out_of_the_asgi_router` (`core/tests/test_ws_session_metrics.py`) reads the routed set from the real `backend.asgi` URL router and asserts it against the declared set. Two properties fall out: a *newly* unrouted consumer fails immediately, and the moment retrospective becomes routed the test fails telling the next developer to remove the exemption, so it can't outlive the fix. I verified that by temporarily adding the route — the series appeared at once and the suite failed with `now routed, remove from KNOWN_UNROUTED_CHANNELS: ['retrospective']` before I reverted it.

Coverage is deliberately read from the router rather than by importing the consumers, because importing one registers its series in the test process — a suite that imports them directly creates the very series it claims to verify. That flaw was in my first version of this suite and is what hid this gap.

**What I've added from your comment:**

- **Dashboard note** — the board description now states that six channels appear and that retrospective's absence means *unreachable, not idle*, so nobody reads the gap as zero traffic.
- **Follow-up** — recorded against [MED-319](https://linear.app/mediajira/issue/MED-319/chat-grafana-dashboard-for-websocket-session-counts) for now, and I'd suggest splitting it into its own issue so it doesn't get closed along with this PR. My recommendation there is deletion rather than repair, given nothing connects to it and nothing broadcasts to it, but that should be a deliberate decision rather than something folded into a metrics PR. Happy to raise it if you agree.

## 3. Label cardinality — agreed, tightened

You're right to gate this. `channel` is bounded, and I've made that explicit rather than leaving it to be trusted.

It has exactly one source: a class-level constant on each consumer.

```
asset/consumers.py:14:          ws_channel = 'asset'
chat/consumers.py:89:           ws_channel = 'chat'
csm/consumers.py:29:            ws_channel = 'csm'
meetings/consumers.py:31:       ws_channel = 'meetings'
portal/consumers.py:20:         ws_channel = 'portal'
retrospective/consumers.py:14:  ws_channel = 'retrospective'
spreadsheet/consumers.py:83:    ws_channel = 'spreadsheet'
```

It is never derived from the connection — not from `scope`, `url_route` kwargs, the user, the room, or the sheet id. The ceiling is the number of consumer classes, so no volume of traffic can mint a series.

Taking your suggestions:

- **Comment explaining the allowed values and why they are safe** — added to `backend/core/metrics.py`, including the specific failure it's guarding against, so the next person doesn't make it dynamic without noticing.
- **Automated test asserting only the expected label set** — added `test_channel_label_set_is_bounded_and_declared`, which pins the seven allowed values and fails on any undeclared series. It also asserts `ws_channel` is a class-level string on every consumer, so making it per-connection breaks the build rather than the monitoring.
- **Plus a runtime check** — `test_sessions_do_not_create_new_label_values` opens sessions on three different URLs and asserts the label set is unchanged. That covers the actual concern (series growing with usage) rather than just the declaration.

I didn't add the Prometheus relabeling rule. With the label set closed at the source, relabeling would map values that can't occur, and it would silently absorb a genuine mistake that I'd rather have fail loudly in CI. Say the word if you'd prefer it as defence in depth.

---

**Summary of changes in this push**

| Change | File |
|---|---|
| Cardinality invariant documented at the gauge | `backend/core/metrics.py` |
| `test_channel_label_set_is_bounded_and_declared` | `backend/core/tests/test_ws_session_metrics.py` |
| `test_sessions_do_not_create_new_label_values` | `backend/core/tests/test_ws_session_metrics.py` |
| `test_two_processes_aggregate_to_the_global_total` | `backend/core/tests/test_ws_session_metrics.py` |
| Coverage note: retrospective is unreachable, not idle | `devops/grafana/websocket-sessions-dashboard.json` |

Suite is **14 passed**. (The PR description said 10 — that predated the ASGI-router test and these three; corrected throughout.)

Not changed, with reasoning above: the in-process gauge, and `RetrospectiveConsumer`'s routing. Glad to take either further if you'd still like them changed.

quanwangniuniu (approved):
Good job

PR #785, opened 2026-08-17
Add audit events and password rotation

qiwangai:
## Summary
- Add centralized append-only AuditEvent storage with HMAC signatures, database mutation protection, and backend emit helpers.
- Emit audit events for key sensitive operations, including project/member changes, password rotation, data export, and account deletion.
- Add elevated-user password rotation tracking, warning/enforcement middleware, frontend reminder/redirect flow, and hardening guards.


quanwangniuniu (changes requested):
1) Audit events may be written before the database transaction commits (lost or duplicated events)
- Why this is a problem
  - If the code emits and signs an AuditEvent during the same transaction but before commit (or outside of a post-commit hook), a subsequent rollback will leave the system state changed without a matching audit entry, or a retry might produce duplicate audit entries. That breaks the audit integrity guarantee (append-only, correct timeline).
- Concrete mitigation / solution
  - Emit audit events only after the application DB transaction successfully commits. Implement a post-commit queue or use DB-level AFTER COMMIT hooks (or an explicit “emit after commit” worker). In practice:
    - Collect audit payloads during the transaction and enqueue them in a per-request post-commit buffer.
    - On successful commit, sign and persist the audit event (or publish to the durable audit store).
    - Ensure the audit persistence is itself durable and idempotent; include a unique event_id (UUID) and a natural monotonic field (timestamp + sequence or DB-generated monotonic id) so retries are idempotent.
  - If you must write audit records inside the same DB, write them in the same transaction but only as "tentative" rows; then flip a committed flag via an atomic commit process, or prefer the post-commit approach so audit == source-of-truth ordering is preserved.

2) HMAC signing and key management look incomplete / risky (no key IDs, possible non-constant comparison, no rotation plan)
- Why this is a problem
  - Without an explicit key identifier (kid) and rotation scheme, verifying services cannot know which key to use; rotating keys may require re-signing or will break verification. If signature verification uses naive equality, it may be vulnerable to timing attacks. Also if the HMAC covers a non-canonical payload, verifiers may compute different bytes and fail to verify.
- Concrete mitigation / solution
  - Use an HMAC envelope with metadata:
    - Include a key id (kid) and algorithm id (alg) in the audit event envelope.
    - Canonicalize the payload deterministically before signing (e.g., stable JSON encoding: sorted keys, consistent whitespace, explicit nulls).
    - Use a constant-time compare when verifying signatures (e.g., crypto.timingSafeEqual or library equivalent).
    - Add a key-rotation strategy:
      - Maintain a key metadata table with active_key_id and versioned keys.
      - When rotating, add new key, make it active, accept signatures from previous N keys for a configurable window, and re-sign only if necessary.
      - Log key rotations as audit events (but do not sign key rotation events with the new key before it’s live).
  - Consider switching to an authenticated signature scheme with key IDs (e.g., HMAC-SHA256 with kid header) or to an asymmetric signature (ed25519 / RSA) if cross-service verification is needed and you want non-repudiation + easier rotation of verification keys.

3) Sensitive data leakage and password-rotation handling (audit contents and session handling)
- Why this is a problem
  - Audit event payloads may accidentally include sensitive fields (passwords, password hashes, auth tokens) or too much PII. Also the password-rotation flow needs to ensure active sessions and tokens are invalidated appropriately; otherwise users who rotated a password could leave elevated sessions open (security gap).
- Concrete mitigation / solution
  - Audit event content policy:
    - Define and enforce an allowlist of fields that may be stored in audit events. Redact or omit any raw secrets (passwords, OTP codes, token strings). If you must record that a value changed, record metadata (e.g., "password_changed": true) rather than the value.
    - Add automatic redaction helpers in the audit emit path and add unit tests that try to emit objects containing password/token fields to ensure they get redacted.
  - Session and token handling on password rotation:
    - On password rotation for elevated users: invalidate all existing sessions and tokens (or at least elevated sessions), and require re-authentication. Specifically:
      - Revoke refresh tokens and force short-lived access tokens to expire (or blacklist them).
      - Clear server-side session records and rotate session cookies/session keys.
      - If using JWTs without server-side revocation, switch to a revocation list / token version stored with the user record and increment token_version on password rotation; have token verification check token_version.
    - Implement a grace/notification UX: inform user of sessions that were invalidated, let admins see recent session activity, and provide a short grace period only where operationally required.
  - Frontend redirect / reminder flow hardening:
    - Validate redirect URLs against an allowlist to prevent open-redirect attacks.
    - Avoid sending raw tokens or PII in querystring parameters during reminder/redirect flows.

Additional pragmatic items (short)
- Add unit and integration tests for:
  - Post-commit emission behavior (simulate rollback and ensure no events persist).
  - Signature generation and verification using multiple keys (rotation window).
  - Redaction logic for PII/secrets.
- Add monitoring / alerting:
  - Track audit delivery failures and signature verification errors, and alert if gaps are detected or if event counts drop unexpectedly.

quanwangniuniu (approved):
Good job!

PR #786, opened 2026-08-17
MED-255: Serialize token refresh to avoid double-login

Popp180:
## Summary
Concurrent 401s in api.ts used to each independently call /auth/token/refresh/,
and one refresh could invalidate another's new token, forcing a logout. A
shared promise now collapses concurrent refreshes into one call.

## Key Changes
- `api.ts` — added a shared promise (`sharedRefreshPromise` / `getSharedRefreshedToken`)
- `authStore.ts` — guarded `initializeAuth()` against re-entrant refresh calls
- `preferencesApi.ts` — reused api.ts's shared promise instead of its own refresh call

## Test plan
- [x] Jest: 4 passed (tokenRefreshSingleFlight, authInitializeReentrantRefresh) + 215/215 full regression
- [x] Playwright: tab-storm-refresh.spec.ts passed on Chromium
- [x] npm run build: 0 errors

## Notes
ROTATE_REFRESH_TOKENS is already False on the backend, so the ticket's
backend grace-period item is a non-issue — not implemented.

Ticket: MED-255


quanwangniuniu (approved):
Good job

PR #787, opened 2026-08-18
Kevinhuangtaowen/med 353 deploy single instance app to google cloud compute engine

07Coder29:
1. Add http redirect and secret management to deployment — Introduces fetch-secrets.sh, a script that pulls secrets (likely from GCP Secret Manager) at deploy time, and wires those secrets into docker-compose.pro.single.yml. Also adds an HTTP→HTTPS redirect rule to nginx/nginx.conf.pro, plus a small fix in backend/agent/gemini_client.py.

2. Implement Github Action listener — Adds ops/webhook/deploy_listener.py, a webhook server that listens for deploy triggers from GitHub Actions and redeploys the stack on the VM, along with recompose.sh to rebuild/restart the docker-compose services. The CI workflow (.github/workflows/mediajira-ci.yml) is updated to call this webhook after a successful build, nginx.conf.pro is extended to route to the listener, and fetch-secrets.sh gains an additional secret needed by it.

### DIFF
--- a/.github/workflows/mediajira-ci.yml
+++ b/.github/workflows/mediajira-ci.yml
@@ -505,9 +505,11 @@ jobs:
       - name: Trigger deploy (main)
         env:
           DEPLOY_WEBHOOK_URL: ${{ secrets.MEDIAJIRA_DEPLOY_WEBHOOK_URL_MAIN }}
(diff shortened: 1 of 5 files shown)


quanwangniuniu (approved):
Good job

PR #788, opened 2026-08-20
refactor(chat): extract consumer subscription registry + MED-303

zizhenliu0427:
# MED-303 — Extract Chat Consumer Subscription Registry

## Summary

This PR extracts chat-group subscription bookkeeping and group dispatch from `ChatConsumer` into a typed `SubscriptionRegistry`.

The registry owns the relationship between one WebSocket channel and its chat groups, including Channel Layer calls, local state, concurrency serialisation, cleanup, and leak instrumentation. Django authentication, database entitlement lookup, connection lifecycle, and the WebSocket protocol remain in the consumer.

QA also exposed a frontend WebSocket lifecycle race: a superseded socket could finish closing after its replacement connected and clear the shared live-socket reference. The included guard ignores that stale close, preserving typing and other outbound events without changing the UI or WebSocket protocol.

Linear ticket: https://linear.app/mediajira/issue/MED-303/chat-extract-chat-consumer-subscription-registry

QA video: [Google Drive](https://drive.google.com/file/d/1opjsQKwnGAcH9Bv8lBw1tpx9tLPq3pko/view?usp=sharing)

## Why

`backend/chat/consumers.py` previously mixed three concerns:

1. WebSocket connection lifecycle and authentication.
2. Chat-group subscription bookkeeping.
3. Presence and message event fan-out.

The bookkeeping could only be exercised through the consumer lifecycle, which made concurrency and failure behaviour difficult to test in isolation. Moving it to a small registry provides one owner for Channel Layer membership and its matching in-memory state.

### Registry boundary

`SubscriptionRegistry` receives only:

- a `ChannelLayerProtocol` implementation;
- the connection's `channel_name`.

It does not import Django models, query the database, authenticate users, or read WebSocket messages. Tests can therefore use an in-memory fake Channel Layer.

The public API is:

| Method | Behaviour |
| --- | --- |
| `add(group_name)` | Idempotently joins one group and reports whether it was added. |
| `remove(group_name)` | Idempotently leaves one group and reports whether it was removed. |
| `sync(allowed)` | Atomically derives the diff, removes revoked groups first, then adds new groups. |
| `snapshot()` | Returns an immutable, consistent `frozenset`. |
| `dispatch(event, groups=None)` | Publishes to the current snapshot or an explicit prior snapshot. |
| `clear()` | Best-effort terminal cleanup, returning subscribed/released/leaked groups. |

`SubscriptionDelta` and `CleanupResult` are frozen dataclasses so callers receive typed, immutable results.

### State consistency and failure behaviour

Channel Layer I/O happens before the corresponding local state mutation. If an operation fails, the registry never claims the subscription change completed.

During `sync()`:

- removals run before additions, so revoked access is removed first;
- completed operations remain represented if a later operation fails;
- the exception propagates to the consumer, which closes a connection whose entitlements are no longer trustworthy.

During `clear()`:

- every tracked group is attempted, even if an earlier discard fails;
- successfully discarded and possibly leaked groups are returned separately;
- the local registry and active gauge are always released because the connection no longer owns them;
- repeated cleanup is harmless and the terminal registry cannot be reopened.

### Thread safety

An `asyncio.Lock` alone is safe only inside one event loop. The registry records its owner loop when constructed and sends calls from foreign OS threads back to that loop through `asyncio.run_coroutine_threadsafe`.

This keeps both the lock and the Channels client on one event loop while allowing independent threads to invoke the public API safely. A dedicated test starts eight OS threads, each with its own event loop, and verifies that all Channel Layer work still executes on the registry owner loop.

### Consumer integration

`ChatConsumer` now delegates chat-group work as follows:

- `connect()` constructs the registry and performs initial sync before `accept()`.
- `sync_chat_groups()` obtains the authoritative allowed chat IDs from the database, converts them to group names, and calls `registry.sync()`.
- membership-change events still call `sync_chat_groups()`; the external event protocol is unchanged.
- chat-group presence fan-out uses `registry.snapshot()` and `registry.dispatch()`.
- `disconnect()` calls `registry.clear()` and retains the returned prior subscriptions for the offline-presence broadcast.
- `ChatConsumer.__call__()` performs a final idempotent cleanup for exception paths where Channels does not invoke `disconnect()`.

The personal `chat_user_<id>` group remains in the consumer because it is connection lifecycle state rather than database-derived chat membership.

## Metrics

| Metric | Type | Meaning |
| --- | --- | --- |
| `chat_active_subscriptions` | Gauge | Live chat-group subscriptions owned by registries in this process. |
| `chat_subscription_leaks_total{reason="discard_failed"}` | Counter | Groups that could not be discarded during terminal cleanup. |

Personal user groups are deliberately excluded from the active gauge. The leak counter uses one bounded reason label; user, channel, and group identifiers are not metric labels.

## Files changed

| File | Change |
| --- | --- |
| `backend/chat/subscriptions.py` | New typed, concurrency-safe registry. |
| `backend/chat/consumers.py` | Delegates chat-group sync, cleanup, snapshot, and dispatch. |
| `backend/chat/metrics.py` | Adds active-subscription and cleanup-leak metrics. |
| `backend/chat/tests/test_subscriptions.py` | Adds isolated unit, failure, concurrency, OS-thread, metric, and crash-cleanup tests. |
| `frontend/src/hooks/useChatWebSocket.ts` | Prevents a stale socket close from clearing its live replacement. |
| `frontend/src/__tests__/hooks/useChatWebSocket.pinAlerts.test.tsx` | Covers the replacement-socket race and confirms outbound typing still uses the live socket. |

Diff against `prod-preview`: **710 insertions, 25 deletions across 6 files**.

### Registry tests

```bash
docker exec backend-dev pytest -o addopts="" --reuse-db chat/tests/test_subscriptions.py -q
```

Result: **12 passed**.

Coverage includes:

- idempotent add/remove;
- correct sync delta and remove-before-add ordering;
- accurate state after partial Channel Layer failure;
- immutable snapshots;
- concurrent add/remove and sync;
- eight real OS threads calling through independent event loops;
- normal and failed cleanup metrics;
- post-clear explicit-snapshot dispatch;
- terminal/idempotent cleanup;
- registry cleanup when the consumer crashes.

### Consumer and realtime regression tests

```bash
docker exec backend-dev pytest -o addopts="" --reuse-db \
  chat/tests/test_consumers.py \
  chat/tests/test_realtime.py \
  core/tests/test_ws_session_metrics.py -q
```

Result: **33 passed**.

### Full chat suite

```bash
docker exec backend-dev pytest -o addopts="" --reuse-db chat/tests/ -q
```

Result: **226 passed**.

### Frontend WebSocket regression test

```bash
docker compose -f docker-compose.dev.yml exec -T frontend \
  npm test -- --runInBand \
  src/__tests__/hooks/useChatWebSocket.pinAlerts.test.tsx
```

Result: **3 passed**. This includes the stale-close/replacement-socket regression case.

After merging the latest `prod-preview`, the registry, consumer, realtime, and WebSocket metric regression set was rerun together: **45 passed**.

### REST + WebSocket E2E

K6 was run with two users against a temporary Daphne process with `CHAT_CHANNEL_GROUPS_ENABLED=True`.

Result:

- **7/7 checks passed**;
- **2/2 WebSocket upgrades succeeded**;
- **2/2 REST messages were accepted and received by every peer**;
- zero connection failures, send failures, ready-after-barrier events, and duplicate deliveries;
- `chat_send_http_ms` p95: **67.56 ms**;
- `chat_ws_delivery_ms` p95: **495.95 ms**;
- while connected: `chat_active_subscriptions 4.0` and `chat_subscription_leaks_total{reason="discard_failed"} 0.0`;
- after disconnect: `chat_active_subscriptions 0.0` and `chat_subscription_leaks_total{reason="discard_failed"} 0.0`.

The active gauge moved from `0.0` before the run, to `4.0` while connected, and back to `0.0` after disconnect. The two prepared users each had two active chat-group memberships, so the gauge correctly counted four connection-to-group subscriptions.

### Static checks

- `git diff --check`: passed.
- Python compilation: passed.
- ESLint for the two changed frontend files: passed.

## Expected result

- Chat-group membership has one testable source of truth per connection.
- Concurrent calls cannot drift local state from Channel Layer state.
- Failed cleanup is observable without leaving the active gauge permanently inflated.
- Normal WebSocket behaviour and protocol remain unchanged.

## Scope and rollout notes

- Backend registry extraction plus a focused frontend WebSocket lifecycle guard discovered during QA; no new UI.
- No migrations.
- No WebSocket consumer protocol changes.
- `CHAT_CHANNEL_GROUPS_ENABLED` remains the existing rollout switch and is off by default. The enabled path was explicitly exercised in E2E.
- Metrics are process-local Prometheus series and should be aggregated across application instances in dashboards.
- A failed Redis group discard is counted as a possible leak; the Channel Layer's group expiry remains the eventual cleanup backstop.

## Review checklist

- [x] Registry API is typed and independently testable.
- [x] Real OS-thread concurrency is covered.
- [x] Consumer delegates chat-group bookkeeping and dispatch.
- [x] Cleanup runs on normal disconnect and unexpected consumer exit.
- [x] Active and leak metrics are exported.
- [x] Chat tests and REST/WebSocket E2E pass locally.
- [x] Frontend stale-close regression is covered; no UI, migration, or protocol changes.
- [x] QA passed and the video link is included above.
- [ ] PR CI passes.


quanwangniuniu (changes requested):
Issue 1 — Blocking / deadlock risk when calling into the owner loop from other threads
Scope: General (applies to the registry's thread-crossing API and tests)

``` 
The registry uses asyncio.run_coroutine_threadsafe(...) to schedule work on the registry's owner loop. If callers call .result() on that Future with no timeout (or the owner loop is stopped), the calling OS thread can block indefinitely and leak resources.

Please make the thread-to-loop boundary robust:
- Always use a bounded wait for run_coroutine_threadsafe results (e.g. future.result(timeout=...) with a configurable timeout).
- On TimeoutError or CancelledError, fail fast: log an informative error, surface a typed RegistryError to the caller, and avoid blocking the thread indefinitely.
- When the registry is closing, cancel outstanding futures / mark the registry terminal so new calls immediately fail rather than block.
- Add a unit test that simulates the owner loop shutting down (or a slow Channel Layer) and asserts that calls from other threads do not block forever but return with a timeout/error.

Suggested small code change sketch:
- Replace direct future.result() with something like:
  future = asyncio.run_coroutine_threadsafe(self._do_add(...), self._owner_loop)
  try:
      return future.result(timeout=self._thread_call_timeout)
  except concurrent.futures.TimeoutError:
      # log & raise a RegistryTimeoutError
      raise RegistryTimeoutError("timed out waiting for registry owner loop")
```

Why this matters
- Without a bounded wait the server thread can block forever if the owner loop stalls or is closed, causing resource leaks and making shutdown/hard-failure handling unreliable. Adding a short, configurable timeout and clear error handling makes the API safe to call from arbitrary threads and avoids subtle production deadlocks.

---

Issue 2 — Dispatch error handling and lack of concurrency control when publishing to many groups
Scope: General (the registry.dispatch implementation and its Channel Layer calls)

``` 
dispatch(event, groups=None) currently fans out to many group names via the Channel Layer but doesn't (a) aggregate or return per-target failures consistently, and (b) limit concurrency when the target set is large. As written this can:
- silently drop or swallow Channel Layer exceptions,
- overload the Channel Layer client when dispatching to many groups at once,
- leave callers uncertain whether some sends failed.

Please improve dispatch by:
- Publishing to a fixed snapshot (already present) but using asyncio.gather(..., return_exceptions=True) so all send operations run and you can collect per-group results.
- Aggregate errors and surface them to the caller as a typed DispatchResult (or raise a summarized exception) so callers can decide whether to close the connection or retry.
- Add an optional concurrency limiter (asyncio.Semaphore) or chunking logic for very large snapshots to avoid overwhelming the Channel Layer.
- Add tests that simulate partial Channel Layer failures and very large group counts to verify that failed sends are reported and do not corrupt registry state.

Suggested pattern:
- Build a list of coroutines for each target send.
- Use an asyncio.Semaphore to cap concurrency:
  async def _bounded_send(sem, coro):
      async with sem:
          return await coro
- results = await asyncio.gather(*[ _bounded_send(sem, send_group(...)) for ... ], return_exceptions=True)
- Inspect results and log/return failures in a DispatchResult.
```

Why this matters
- Dispatch is a cross-cutting behavior: production loads may fan out to thousands of groups. Without backpressure/concurrency limits you risk overwhelming the Channel Layer or hitting connection limits. Without aggregated error reporting, higher layers (the consumer) can't reliably decide whether to close the socket or retry, which undermines the robustness guarantees you described.

---

zizhenliu0427:
Thanks for flagging these issues. They have been addressed in:

**Commit:** `fix(chat): bound registry calls and dispatch fanout + MED-303`  
**SHA:** `12b42ddae`

1. **Owner-loop timeout and shutdown handling**

   Cross-thread calls now have a configurable bounded timeout and raise typed registry errors on timeout, cancellation, closure, or an unavailable owner loop. Closing the registry also cancels outstanding cross-thread work and immediately rejects new mutations. Tests cover a stalled owner loop and shutdown cancellation.

2. **Bounded dispatch and partial failure reporting**

   Dispatch now uses a fixed group snapshot, bounded concurrency, and `asyncio.gather(..., return_exceptions=True)`. A typed `DispatchResult` reports successful and failed groups without changing registry state, and the consumer logs partial failures. Tests cover partial Channel Layer failures and a 100-group fan-out with a verified concurrency cap.

The registry, complete chat suite, and local backend regression tests have been run successfully.

quanwangniuniu (approved):
Good job

PR #790, opened 2026-08-20
Feature/med 344 admin action diff viewer

RainF0rester:
# Ticket Resolution Report — MED-344: Admin Action Diff Viewer

---

## 0. Summary

- Introduces a new `audit` Django app with an `AdminAuditEvent` model that logs all administrative actions (14 action types) to the public schema, covering roles, permissions, projects, and organization management.
- Adds `capture_snapshot()` and `record_audit_entry()` utilities to record before/after state diffs; all calls are wrapped in `transaction.atomic()` to prevent orphaned audit records on failure.
- Wires audit logging into 13 existing views across `access_control` and `core`, and exposes audit data via `GET /api/audit/events/` with filtering by action, project, and target type.
- Adds a frontend audit log page at `/admin/audit-log` with expandable diff rows (red/green field-level changes), relative timestamps, action filtering, and pagination (20/page).
- Adds an `AuditCard` component on the Overview page showing the 6 most recent events.
- Uses `db_constraint=False` on the `project` FK to handle cross-schema FK limitations in the multi-tenant PostgreSQL setup.

## 1. Ticket Overview

| Field       | Detail                                          |
|-------------|-------------------------------------------------|
| Ticket ID   | MED-344                                         |
| Title       | Admin Action Diff Viewer                        |
| Assignee    | yulin                                           |
| Branch      | `feature/MED-344-admin-action-diff-viewer`      |

---

### Background

Before this ticket, there was no mechanism to track administrative actions in the system. When a role was misconfigured, permissions were changed unexpectedly, or a project was deleted, there was no audit trail to determine who did it, when, or what the state was before the change.

### Impact

- Administrators had no visibility into permission changes or org-level operations.
- Debugging access control issues required inspecting the database directly.
- No before/after comparison was available for any administrative action.

### Related Modules

- `access_control` — role and permission management
- `core` — project, organization, and org admin management
- `audit` — newly created app

---

### New App: `audit`

A new Django app `audit` was created with the following components:

#### `AdminAuditEvent` Model (`audit/models.py`)

Stores one record per administrative action in the **public schema**, making it accessible across all tenants.

| Field         | Type                  | Notes                                                                 |
|---------------|-----------------------|-----------------------------------------------------------------------|
| `id`          | UUIDField (PK)        | Auto-generated, not editable                                          |
| `organization`| ForeignKey (nullable) | The org this action belongs to. SET_NULL on org deletion.             |
| `project`     | ForeignKey (nullable) | The project this action belongs to. `db_constraint=False` — see migrations section. |
| `actor`       | ForeignKey (nullable) | The user who performed the action. SET_NULL if user is deleted.       |
| `action`      | CharField             | One of 14 predefined action codes (see below). Indexed.              |
| `target_type` | CharField             | The model class name of the affected entity (e.g., `"Role"`, `"User"`). |
| `target_id`   | CharField             | Primary key of the affected entity as a string.                       |
| `target_name` | CharField             | Human-readable name of the affected entity.                           |
| `before`      | JSONField (nullable)  | Snapshot of the entity before the action. `None` for create actions.  |
| `after`       | JSONField (nullable)  | Snapshot of the entity after the action. `None` for delete actions.   |
| `timestamp`   | DateTimeField         | Auto-set on creation. Indexed. Default ordering: newest first.        |

**Supported action codes (14 total):**

| Code                       | Description                     |
|----------------------------|---------------------------------|
| `role.created`             | A new role was created          |
| `role.updated`             | A role's name/level was changed |
| `role.deleted`             | A role was deleted              |
| `role.permissions_updated` | Permissions on a role changed   |
| `role.permissions_copied`  | Permissions were copied to a role |
| `user_role.assigned`       | A role was assigned to a user   |
| `user_role.removed`        | A role was removed from a user  |
| `project.updated`          | Project settings were changed   |
| `project.deleted`          | A project was deleted           |
| `project.labels_updated`   | Decision topic labels changed   |
| `org.slug_updated`         | Organization URL slug changed   |
| `org.deleted`              | An organization was deleted     |
| `org.admin_assigned`       | A user was granted org admin    |
| `org.admin_removed`        | Org admin was revoked from user |

---

#### Utility Functions (`audit/utils.py`)

**`capture_snapshot(instance) -> Dict`**

Serializes a Django model instance to a JSON-safe dictionary for use as a `before`/`after` snapshot.

- Uses `model_to_dict()` to capture all editable fields.
- Explicitly adds `id`, `created_at`, `updated_at` (not returned by `model_to_dict`).
- Converts UUID and datetime values to strings for JSON compatibility.

**`record_audit_entry(actor, action, target, before, after, organization, project) -> AdminAuditEvent`**

Inserts a single audit record into the `audit_adminauditevent` table.

- Validates `action` against the 14 predefined codes; raises `ValueError` if invalid.
- Extracts `target_type` from `target.__class__.__name__` and `target_id` from `str(target.pk)`.
- Resolves `target_name` via `_resolve_target_name()`: for User objects, prefers `get_full_name()` → `username` → `email`; for all other models, uses the `name` attribute.
- **Must be called inside a `transaction.atomic()` block** so that the audit entry rolls back together with the main operation if anything fails.

---

#### API Endpoint (`audit/views.py`, `audit/urls.py`)

**`GET /api/audit/events/`**

Returns a paginated list of audit events for the current user's organization.

- **Authentication:** Required (`IsAuthenticated`)
- **Org isolation:** Results are always scoped to the user's `current_organization_id`. Non-staff users with no resolvable org receive an empty result set. Staff/superusers can see all events.

**Supported query parameters:**

| Parameter     | Description                                                                 |
|---------------|-----------------------------------------------------------------------------|
| `action`      | Filter by action code (e.g., `?action=role.updated`)                       |
| `project_id`  | Filter to events for a project **plus** all org-level events (project=None) |
| `target_type` | Filter by model type (e.g., `?target_type=Role`)                            |

**Response fields:** `id`, `organization_id`, `project_id`, `actor_id`, `actor_email`, `actor_name`, `action`, `target_type`, `target_id`, `target_name`, `before`, `after`, `timestamp`

---

### Modified Existing Views

`record_audit_entry()` calls were added to the following views, all wrapped in `transaction.atomic()`:

| File                         | View / Method                     | Action recorded              |
|------------------------------|-----------------------------------|------------------------------|
| `access_control/views.py`    | `roles_list` (POST)               | `role.created`               |
| `access_control/views.py`    | `role_detail` (PUT)               | `role.updated`               |
| `access_control/views.py`    | `role_detail` (DELETE)            | `role.deleted`               |
| `access_control/views.py`    | `update_role_permissions`         | `role.permissions_updated`   |
| `access_control/views.py`    | `copy_role_permissions`           | `role.permissions_copied`    |
| `access_control/views.py`    | `assign_user_role`                | `user_role.assigned`         |
| `access_control/views.py`    | `remove_user_role`                | `user_role.removed`          |
| `core/views.py`              | `ProjectViewSet.perform_update`   | `project.updated`            |
| `core/views.py`              | `ProjectViewSet.perform_destroy`  | `project.deleted`            |
| `core/views.py`              | `ProjectViewSet.decision_topic_label` | `project.labels_updated` |
| `core/views.py`              | `UpdateOrganizationSlugView`      | `org.slug_updated`           |
| `core/admin_views.py`        | `AdminOrganizationViewSet.assign_admin` | `org.admin_assigned`   |
| `core/admin_views.py`        | `AdminOrganizationViewSet.remove_admin` | `org.admin_removed`    |

*(Note: `org.deleted` is wired in `OrganizationDetailView` but requires superuser; not exposed in normal admin UI.)*

---

### Database Migrations

**`0001_initial`** — Creates the `audit_adminauditevent` table with core fields: `id`, `actor`, `action`, `target_type`, `target_id`, `target_name`, `before`, `after`, `timestamp`.

**`0002_adminauditevent_organization_adminauditevent_project_and_more`** — Adds `organization` and `project` ForeignKey fields. Also renames `actor`'s `related_name` from `admin_audit_events` to `audit_events_acted` to avoid name collision with the new `organization` FK.

**`0003_adminauditevent_project_no_db_constraint`** — Sets `db_constraint=False` on the `project` FK. This is required because `core_project` lives in the tenant schema (e.g., `org_big`) while `audit_adminauditevent` lives in the `public` schema. PostgreSQL does not support enforced FK constraints across schemas, so the constraint is removed while the logical reference is kept.

---

### Frontend

**New page: `/admin/audit-log` (`frontend/src/app/(project)/admin/audit-log/page.tsx`)**

- Displays all audit events in a table with columns: **Type badge** / **Actor** / **Action** / **Target** / **Time**
- Time is shown as relative time (e.g., `2h ago`); hover shows the full timestamp.
- Rows with `before`/`after` data are expandable — clicking shows an inline diff view with field-level changes highlighted in red (removed) and green (added).
- Supports filtering by action type via a dropdown; resets to page 1 on filter change.
- Paginated at 20 events per page.

**New component: `AuditCard` (`frontend/src/components/overview/AuditCard.tsx`)**

- Displays the 6 most recent audit events on the Overview page.
- "View all actions" link navigates to the full audit log page.

**New types: `frontend/src/types/audit.ts`**

Added `AdminAuditAction`, `AdminAuditEvent`, `PaginatedAdminAuditEvents`, and `AdminAuditEventFilters` types.

**Nginx config (`nginx/nginx.conf`):**

Added `audit-log` to the frontend admin routes regex so `/admin/audit-log` is proxied to the Next.js frontend, not the Django admin.

---

### Automated Tests

**`audit/tests/test_utils.py` — 7 test cases**

| Test | What it verifies |
|------|-----------------|
| `test_returns_dict` | `capture_snapshot` returns a dict |
| `test_contains_name_and_level` | Editable fields are included in snapshot |
| `test_id_is_string` | UUID `id` is converted to string |
| `test_created_at_is_string` | `created_at` datetime is converted to string |
| `test_snapshot_is_independent_copy` | Snapshot is not affected by later model changes |
| `test_creates_audit_event` | `record_audit_entry` creates exactly one DB row |
| `test_event_fields_are_correct` | All fields stored correctly in DB |
| `test_create_action_before_is_none` | `before=None` for create events |
| `test_delete_action_after_is_none` | `after=None` for delete events |
| `test_invalid_action_raises_value_error` | Unknown action codes raise `ValueError` |
| `test_rolls_back_with_transaction` | Audit entry rolls back if outer transaction fails |
| `test_actor_can_be_none` | `actor=None` is accepted for system-triggered events |

**`audit/tests/test_views.py` — 15 test cases**

| Test | What it verifies |
|------|-----------------|
| `test_unauthenticated_returns_401` | Unauthenticated requests rejected |
| `test_authenticated_returns_200` | Authenticated requests succeed |
| `test_empty_list` | Zero results when no events exist |
| `test_returns_created_events` | Created events appear in response |
| `test_response_contains_expected_fields` | All required fields present |
| `test_organization_id_and_project_id_in_response` | org_id and project_id included |
| `test_only_returns_current_org_events` | Events from other orgs are never returned |
| `test_filter_by_action` | `?action=` parameter filters correctly |
| `test_filter_by_target_type` | `?target_type=` parameter filters correctly |
| `test_filter_by_unknown_action_returns_empty` | Unknown action returns 0 results |
| `test_filter_by_project_id_returns_project_events` | Project-scoped events returned |
| `test_filter_by_project_id_includes_org_level_events` | Org-level events included with `?project_id=` |
| `test_filter_by_project_id_excludes_other_projects` | Other projects' events excluded |
| `test_no_project_filter_returns_all_org_events` | All org events returned without filter |
| `test_events_ordered_newest_first` | Results ordered by timestamp descending |

### Manual Test Scenarios

| Scenario | Expected Result |
|----------|----------------|
| Create a new role | Log entry: "Created a new role", target = role name |
| Rename a role | Log entry: "Renamed a role", diff shows old/new name |
| Delete a role | Log entry: "Deleted a role", no expandable diff |
| Update role permissions | Log entry: "Changed role permissions", diff shows permission changes |
| Copy permissions to another role | Log entry: "Copied permissions to role", diff shows permissions |
| Assign a role to a member | Log entry: "Assigned a role to member", target = username |
| Remove a role from a member | Log entry: "Removed a role from member", target = username |
| Update project settings | Log entry: "Updated project settings", diff shows changed fields |
| Delete a project | Log entry: "Deleted a project", no expandable diff |
| Update decision labels | Log entry: "Updated project labels", diff shows topic/title |
| Grant org admin to a user | Log entry: "Granted admin access to member", target = username |
| Revoke org admin from a user | Log entry: "Revoked admin access from member", target = username |
| Filter by action type | Only matching records are shown |
| Expand a diff row | Before/after field changes shown in red/green |
| Check org isolation | User from org A cannot see org B's audit events |

---

## 5. Lessons Learned

- **Cross-schema FK constraints** — In a multi-tenant PostgreSQL setup where the audit table lives in the `public` schema and `Project` lives in a tenant schema, enforced FK constraints are not possible. `db_constraint=False` must be set on the `project` FK field. This was discovered when the initial migration failed in the test environment.

- **`model_to_dict` omits non-editable fields** — `id`, `created_at`, and `updated_at` are read-only and excluded from `model_to_dict` output. These must be manually added in `capture_snapshot()` for the diff to be meaningful.

- **User objects have no `name` field** — Django's `AUTH_USER_MODEL` uses `username` and `email`, not `name`. The generic `getattr(target, 'name', str(target.pk))` fallback returned the user's PK (e.g., `"7"`) instead of their username. A dedicated `_resolve_target_name()` helper was added to handle User objects correctly.

- **`transaction.atomic()` is required** — `record_audit_entry()` must be called inside an atomic block so the audit entry is rolled back if the main operation fails. Calling it outside the transaction would leave orphan audit records for failed operations.

---

## 6. References

- Jira Ticket: MED-344
- Branch: `feature/MED-344-admin-action-diff-viewer`
- Key files:
  - `backend/audit/models.py`
  - `backend/audit/utils.py`
  - `backend/audit/views.py`
  - `backend/audit/serializers.py`
  - `backend/audit/tests/`
  - `frontend/src/app/(project)/admin/audit-log/page.tsx`
  - `frontend/src/components/overview/AuditCard.tsx`


quanwangniuniu (changes requested):
I'll load the PR understanding ability and analyze this pull request for issues.Now let me fetch the PR's changed files to identify the out-of-scope changes:Now let me get the specific diff that was shown in the URL to examine it more closely:Based on my review of PR #790, here are **7 key issues** I identified:

---

## **Issue 1: Sidebar Access Control Logic Bug**
**Problem:** The `Sidebar.tsx` file shows that the effective role level correctly distinguishes between admins and regular users:
```typescript
const effectiveRoleLevel = (authUser?.is_staff || authUser?.is_org_admin) ? 1 : userRoleLevel;
```
However, this logic is **not consistently applied** when the component is initially rendered if the auth store hasn't hydrated yet.

**Why it's bad:** On page load before Zustand hydration, `authUser` will be `null`, so `effectiveRoleLevel` defaults to `userRoleLevel` (likely 10), causing staff/org admins to see the sidebar without admin menu items until hydration occurs. This creates a flickering/race condition.

**Solution:** Immediately check localStorage for the persisted auth state on mount and use that as a fallback:
```typescript
useEffect(() => {
  const storedAuth = readPersistedAuthState()?.state?.user;
  if (storedAuth?.is_staff || storedAuth?.is_org_admin) {
    // Preload admin items
  }
}, []);
```

---

## **Issue 2: Cross-Schema FK Missing Validation on Project**
**Problem:** The `AdminAuditEvent.project` field uses `db_constraint=False` to bypass cross-schema foreign key constraints:
```python
project = models.ForeignKey(
    'core.Project',
    on_delete=models.SET_NULL,
    db_constraint=False,  # Project lives in tenant schema
)
```
Without database constraints, **invalid/orphaned project IDs** can be stored in the audit log.

**Why it's bad:**  
1. Data integrity is lost — you can write any integer to `project_id`, even if it doesn't exist in any tenant
2. Audit queries filtering by project become unreliable (no FK guarantee)
3. No database-level protection against typos in `record_audit_entry(project=...)`

**Solution:** Add application-level validation in `record_audit_entry()`:
```python
if project is not None:
    if not Project.objects.filter(pk=project.pk).exists():
        raise ValueError(f"Project {project.pk} does not exist in any tenant")
```

---

## **Issue 3: Migration Dependency Chain Fragility**
**Problem:** There are **3 migration files** in `backend/audit/migrations/`:
- `0001_initial.py` — defines AdminAuditEvent without org/project FK
- `0002_*_organization_adminauditevent_project.py` — adds org and project FK
- `0003_*_project_no_db_constraint.py` — removes DB constraint from project FK

This chain is **brittle and easy to break** during deployment:
```python
dependencies = [
    ('core', '0014_dataexportrequest'),  # Hardcoded core version
    ('audit', '0002_adminauditevent_organization_adminauditevent_project_and_more'),
]
```

**Why it's bad:**  
- If someone creates a new core migration (becomes `0015_*`), this dependency breaks
- Requires 3 schema updates instead of 1, increasing deployment risk
- No rollback path if migration 3 fails after migration 2 succeeds

**Solution:** Combine all three migrations into a single `0001_initial.py`:
```python
class Migration(migrations.Migration):
    initial = True
    dependencies = [
        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
        ('core', '0014_dataexportrequest'),
    ]
    operations = [
        # Create AdminAuditEvent with all fields including:
        # - organization FK (SET_NULL)
        # - project FK (SET_NULL, db_constraint=False)
        # - All other fields
    ]
```

---

## **Issue 4: Missing Django App Registry in `audit/__init__.py`**
**Problem:** The `backend/audit/__init__.py` file is **empty**:
```python
# frontend/src/audit/__init__.py
(empty file)
```
Meanwhile, `settings.py` registers:
```python
INSTALLED_APPS = [
    ...
    'audit.apps.AuditConfig',  # Expects __init__ to export or configure it
    ...
]
```

**Why it's bad:** While this technically works (Django doesn't strictly require app config in `__init__.py`), it's a **common error pattern** that breaks when developers add signals, ready hooks, or autodiscovery later.

**Solution:** Add an explicit app config import:
```python
# backend/audit/__init__.py
from django.apps import AppConfig

__all__ = ['AuditConfig']
```

---

## **Issue 5: Sidebar Navigation Item Doesn't Check User Is Authenticated**
**Problem:** The audit log link is added to the sidebar conditionally:
```typescript
if (userRoleLevel && userRoleLevel <= 2) {
  children: [
    { name: "Audit Log", href: "/admin/audit-log", icon: ScrollText }
  ]
}
```
But there's **no check** that the user is actually logged in (`authUser` exists).

**Why it's bad:** If `userRoleLevel=2` is set as a default prop but the user isn't authenticated, the Audit Log link appears in the sidebar. Clicking it will redirect to login, confusing users.

**Solution:**
```typescript
if (authUser && userRoleLevel && userRoleLevel <= 2) {
  // Add audit log to children
}
```

---

## **Issue 6: Frontend Types and API Mismatch**
**Problem:** The frontend defines `AdminAuditEvent` with typed `before` and `after`:
```typescript
// frontend/src/types/audit.ts
export interface AdminAuditEvent {
  before: Record<string, unknown> | null;
  after: Record<string, unknown> | null;
}
```
But the backend's `AdminAuditEventSerializer` returns them as-is from the `JSONField`:
```python
# backend/audit/serializers.py
fields = [
    "before",  # JSONField, no transformation
    "after",   # JSONField, no transformation
]
```
If **non-serializable objects** (e.g., datetime, UUID) make it into `before`/`after`, the API response will fail.

**Why it's bad:** The `capture_snapshot()` function tries to convert fields to strings:
```python
data[field] = str(value) if value is not None else None
```
But it only handles `id`, `created_at`, `updated_at`. If a role/project has a DateTimeField or UUIDField not explicitly converted, the JSON encoder will fail.

**Solution:** Ensure all non-JSON-serializable fields are explicitly converted in `capture_snapshot()`:
```python
def capture_snapshot(instance) -> Dict[str, Any]:
    data = model_to_dict(instance)
    
    # Convert all special types
    for field in data:
        if hasattr(data[field], 'isoformat'):
            data[field] = data[field].isoformat()  # datetime, date, time
        elif hasattr(data[field], 'hex'):
            data[field] = str(data[field])  # UUID
    return data
```

---

## **Issue 7: Nginx Configuration Regex Doesn't Include Trailing Slash Variants**
**Problem:** The nginx config adds `audit-log` to the next.js allowlist:
```nginx
location ~ ^/admin/(roles|permissions|approvers|ticket-forms|csm/settings|audit-log|override-audits) {
    proxy_pass http://$frontend_host$request_uri;
}
```
But the regex **doesn't handle trailing slashes or sub-routes**:
```
✗ /admin/audit-log/      — doesn't match (trailing slash)
✗ /admin/audit-log?page=2 — matches but might cause issues with trailing slash
```

**Why it's bad:** Next.js app router treats `/admin/audit-log` and `/admin/audit-log/` as separate routes. If the frontend sends requests with trailing slashes, they'll hit Django's `/admin/` instead of Next.js, causing 404s.

**Solution:** Update the regex to match trailing slashes:
```nginx
location ~ ^/admin/(roles|permissions|approvers|ticket-forms|csm/settings|audit-log|override-audits)/?$ {
    proxy_pass http://$frontend_host$request_uri;
}
```

---

### **Summary Table**

| # | Issue | Severity | Component | Fix Effort |
|---|-------|----------|-----------|-----------|
| 1 | Sidebar auth hydration race condition | Medium | Frontend | 1h |
| 2 | Missing FK validation on project ID | High | Backend | 2h |
| 3 | Brittle 3-migration chain | Medium | Backend | 2h |
| 4 | Empty audit app `__init__.py` | Low | Backend | 15m |
| 5 | No auth check on sidebar link | Medium | Frontend | 15m |
| 6 | Non-serializable fields in snapshots | High | Backend | 1h |
| 7 | Nginx regex missing trailing slash | Medium | Infra | 15m |

quanwangniuniu (approved):
Approved!

PR #791, opened 2026-08-21
MED-253: Keep notification SSE alive past Nginx idle timeout

JieyaZ:
Preserves the existing 25-second SSE heartbeat, configures Daphne for long-lived HTTP streams, adds frontend reconnection and connection health tracking, and implements UUID-based missed-notification replay. Includes backend, frontend, and Playwright test coverage. No database migrations.


quanwangniuniu (changes requested):
## **Issue 1: Race Condition in Notification Replay Logic**

**Why it needs to be dealt with:**
The current implementation has a subtle but dangerous race condition:

```python
# Line 70-80: Query cursor notification
cursor = Notification.objects.filter(recipient_id=user_id, pk=last_event_id).only("id", "created_at").first()

# Then query all later notifications
qs = Notification.objects.filter(recipient_id=user_id).filter(
    Q(created_at__gt=cursor.created_at)
    | Q(created_at=cursor.created_at, id__gt=cursor.id)
).order_by("created_at", "id")
```

**The problem:** If a notification is created *between* these two queries with the same `created_at` timestamp as the cursor, the tie-breaking logic `id__gt=cursor.id` could miss notifications. Additionally, this is vulnerable to notifications being deleted or the cursor ID becoming invalid after the first query, leading to silent data loss.

**Potential solutions:**
1. **Use a transaction with READ COMMITTED isolation** to ensure the cursor lookup and range query are atomic
2. **Add constraint checks** to verify `cursor is not None` before the range query, and wrap the entire operation in a database transaction
3. **Use a snapshot timestamp** that's guaranteed not to change between queries:
   ```python
   def _serialize_missed_notifications(user_id: int, last_event_id: str):
       cursor = Notification.objects.filter(recipient_id=user_id, pk=last_event_id).only("id", "created_at").first()
       if cursor is None:
           return []
       
       # Query using ID comparison alone if timestamps are identical
       qs = Notification.objects.filter(recipient_id=user_id, id__gt=last_event_id).order_by("id")
       return NotificationSerializer(qs, many=True).data
   ```

---

## **Issue 2: Memory Leak - Unbounded `replayed_ids` Set**

**Why it needs to be dealt with:**
On line 161 and 216, the code maintains an in-memory `replayed_ids` set to deduplicate notifications:

```python
replayed_ids: set[str] = set()
# ... add to it at line 179
if not event_id or event_id in replayed_ids:
    continue
```

**The problem:** For users who maintain long-lived SSE connections (the entire goal of this PR), the `replayed_ids` set will grow unbounded with every notification they receive. A user with a browser tab open for weeks could accumulate thousands of UUIDs in memory, consuming RAM and potentially causing performance degradation. There is **no cleanup mechanism** to prune old entries.

**Potential solutions:**
1. **Use a time-bounded LRU cache** instead of a set:
   ```python
   from collections import OrderedDict
   
   # At line 161
   replayed_ids = OrderedDict()  # Track (id, timestamp) pairs
   
   # At line 216
   if event_id in replayed_ids:
       continue
   replayed_ids[event_id] = time.time()
   
   # Periodically clean old entries (older than 5 minutes)
   if len(replayed_ids) > 1000:
       cutoff = time.time() - 300
       replayed_ids = OrderedDict((k, v) for k, v in replayed_ids.items() if v > cutoff)
   ```

2. **Use Redis instead of in-memory storage** for the dedupe set:
   ```python
   # Store replayed notification IDs in Redis with expiration
   dedup_key = f"sse_dedup_{user_id}_{connection_id}"
   r.sadd(dedup_key, event_id)
   r.expire(dedup_key, 300)  # 5-minute TTL
   ```

3. **Limit the set size with a sliding window**:
   ```python
   if len(replayed_ids) > 10000:
       # Keep only the most recent 5000
       replayed_ids = {id for id in sorted(replayed_ids)[-5000:]}
   ```

---

JieyaZ:
### Issue 1 — Replay race condition and deleted cursor

The Redis subscription is still established before the database replay, so notifications created during the replay query are queued by Redis. If an event is returned by both the database and Redis, `replayed_ids` removes the duplicate.

For the actual missing case—the cursor notification being deleted before reconnection—I added a fallback:

- UUID remains the primary cursor.
- The frontend also sends the last event’s `created_at`.
- If the UUID no longer exists, the backend replays from that timestamp inclusively.
- Results remain scoped to the authenticated user and ordered by `(created_at, id)`.
- The frontend suppresses any repeated UUIDs at the timestamp boundary.

### Issue 2 — Replay resource usage

The initial database replay is now bounded to 500 notifications per batch. If more events remain, the stream closes and reconnects using the latest UUID for the next batch. Therefore, the server-side `replayed_ids` set is also bounded by the batch size.

I also added:

- A Redis rolling rate limit: 5 replay attempts per user every 30 seconds.
- HTTP `429` with `Retry-After` when the limit is reached.
- Exponential reconnect backoff with ±20% jitter.
- Environment settings so the batch size, limit and 30-second window can be adjusted after load testing.

Verification completed:

- Backend SSE tests: 30/30 passed
- Frontend SSE/store tests: 27/27 passed
- Frontend production build: passed

quanwangniuniu (approved):
Amazing!

PR #794, opened 2026-08-28
Kevinhuangtaowen/med 141 agent 2 spreadsheet agent integration

07Coder29:
Users can analyze spreadsheets with the Agent and move seamlessly between the Spreadsheet module and the Agent. The Agent uses the existing spreadsheet data and APIs; the Spreadsheet module does not take on Agent-specific logic.


quanwangniuniu (changes requested):
---

### **1. Task Creation Logic Regression – Breaking Change in Behavior**

**Problem:**
- Lines in `backend/agent/executors.py` remove the anomaly confirmation gate entirely
- Previously: `if had_anomalies and not analysis.get('anomalies_confirmed'): → error`
- Now: Tasks are created *regardless* of anomaly confirmation state
- This breaks the workflow for users who expect to confirm/review anomalies before automatic task creation

**Why Fix:**
- Users lose the safety check to review detected anomalies before committing to tasks
- Silent data quality issues may propagate into downstream tasks

**Fix:**
```python
# In create_tasks_from_analysis, restore the gate:
had_anomalies = bool(analysis.get('anomalies'))
if had_anomalies and not analysis.get('anomalies_confirmed'):
    yield {
        "type": "error",
        "content": "Anomalies must be confirmed before creating tasks.",
    }
    return
```

---

### **2. Spreadsheet Import Without Proper Error Handling – Silent Failures**

**Problem:**
- `backend/agent/views.py` (lines 516–535): spreadsheet import wrapped in bare `except Exception: logger.exception(...)` with no recovery
- If import fails, the file is still marked uploaded but spreadsheet creation silently fails
- User gets no feedback that the import didn't work

**Why Fix:**
- Orphaned uploaded files without corresponding spreadsheets waste storage
- Users think data is imported when it isn't

**Fix:**
```python
try:
    spreadsheet_import = create_spreadsheet_from_upload(...)
    ImportedCSVFile.objects.filter(id=result['id']).update(
        spreadsheet_id=spreadsheet_import['spreadsheet_id']
    )
except Exception as e:
    logger.exception("Spreadsheet import failed for file_id=%s", result['id'])
    # Emit error event to frontend; don't let silent failure pass
    yield {"type": "error", "content": f"Import failed: {str(e)[:200]}"}
    return
```

---

### **3. Missing Validation on Gemini Spreadsheet Insights Response**

**Problem:**
- `_normalize_spreadsheet_insights_result()` in `backend/agent/services.py` (lines 478–544) doesn't validate:
  - Empty `summary` strings (could be all whitespace)
  - Anomaly locations with out-of-bounds row/col indices
  - Missing required fields in `recommended_tasks`

**Why Fix:**
- Malformed Gemini responses corrupt the analysis result and cause downstream errors
- Crashes in anomaly rendering if locations point to non-existent rows/columns

**Fix:**
```python
def _normalize_spreadsheet_insights_result(raw, sheet_id=None):
    if not isinstance(raw, dict):
        raise ValueError('Insights response must be a JSON object')
    
    summary = str(raw.get('summary') or '').strip()
    if not summary:
        raise ValueError('Gemini summary is empty or missing')
    
    # Validate location bounds
    for anomaly in raw.get('anomalies', []):
        for loc in anomaly.get('locations', []):
            if not isinstance(loc.get('row'), int) or not isinstance(loc.get('col'), int):
                raise ValueError(f'Invalid location indices in anomaly {anomaly.get("id")}')
```

---

### **4. Sheet Context Not Persisted Across Refresh in Frontend**

**Problem:**
- `AgentChatPage.tsx` stores spreadsheet context in `sessionStorage` under key `"agent-session-spreadsheet-context"`
- But when the page reloads or session is destroyed, the workflow banner context is lost
- User returns to a generic chat session instead of the spreadsheet analysis session

**Why Fix:**
- Breaks the mental model of "continuing a spreadsheet analysis session"
- Forces users to re-upload or re-navigate back to the sheet

**Fix:**
```typescript
// In AgentChatPage, ensure we restore from sessionStorage on mount:
const [sessionSpreadsheetContext, setSessionSpreadsheetContext] = useState<SheetWorkContext | null>(() => {
  if (pendingSpreadsheetPreload) {
    return sheetWorkContextFromPreload(pendingSpreadsheetPreload)
  }
  return readStoredSheetWorkContext()  // <-- This is already done, but ensure it survives reload
})

// Also restore the banner if the session was restored:
useEffect(() => {
  const restoredContext = readStoredSheetWorkContext()
  if (restoredContext) {
    setSessionSpreadsheetContext(restoredContext)
  }
}, [])
```

---

### **5. NL→PatternSteps API Doesn't Validate Column References**

**Problem:**
- `backend/spreadsheet/nl_pattern_service.py` (line 72–78): column validation only checks if the index is in range
- Doesn't verify that referenced columns in the instruction (e.g., "column D") actually exist in the schema
- If Gemini hallucinates a column name, the error message is vague

**Why Fix:**
- Users get cryptic "column index X is out of range" instead of "Column D doesn't exist in your sheet"
- No clear feedback on which instruction part failed

**Fix:**
```python
def _validate_step(step: dict, position: int, sheet_schema: dict) -> dict:
    # ... existing code ...
    
    if step_type == "APPLY_FORMULA":
        target = params["target"]
        col_index = target.get("col")
        
        # Verify column exists in schema
        schema_cols = sheet_schema.get("columns", [])
        if col_index < 0 or col_index >= len(schema_cols):
            col_name = next(
                (c.get("name") for c in schema_cols if c.get("index") == col_index),
                f"Column {col_index}"
            )
            raise ValueError(
                f"Step {position}: APPLY_FORMULA references column '{col_name}' "
                f"which does not exist. Available: {[c['name'] for c in schema_cols]}"
            )
```

---

### **6. Inconsistent "Spreadsheet Insights" Workflow – No Transition Path**

**Problem:**
- PR adds `analyze_spreadsheet_insights` action (lightweight in-sheet analysis)
- But removes the confirmation gate for anomalies in `create_tasks_from_analysis()`
- Pattern: insights auto-confirm anomalies, but the broader agent workflow still requires confirmation
- This creates two conflicting mental models for the same analysis flow

**Why Fix:**
- Confuses users: sometimes tasks auto-create, sometimes they need approval
- Breaks consistency in approval workflows

**Fix:**
```python
# In create_tasks_from_analysis, preserve the gate only for non-insights flows:
analysis_source = workflow_run.data.get("_source")  # Track source in workflow_run
had_anomalies = bool(analysis.get('anomalies'))
is_insights_flow = analysis_source == "spreadsheet_insights"

if had_anomalies and not is_insights_flow and not analysis.get('anomalies_confirmed'):
    yield {
        "type": "error",
        "content": "Anomalies must be confirmed before creating tasks.",
    }
    return
```

---

### **7. Frontend Block IDs Mismatch – Message Reveal Queue Bypassed**

**Problem:**
- `MessageList.tsx`: Pattern & pivot generation modes set `bypassRevealQueue={true}` (line 177)
- But `agentMessageBoardBlockIds.ts` still tries to compute block IDs based on message type
- This causes inconsistency: some messages render immediately, others are stuck in the queue

**Why Fix:**
- UI flickering or messages appearing out of order in pattern/pivot modes
- The reveal animation queue gets confused about which messages to show

**Fix:**
```typescript
// In getMessageBoardBlockIds, add early return for bypassed modes:
export function getMessageBoardBlockIds(
  messages: ChatMessage[],
  options: MessageBoardOptions
): string[] {
  if (options.bypassRevealQueue) {
    // Return empty array; bypass clients handle rendering directly
    return []
  }
  
  // Continue with existing logic...
}
```

---

quanwangniuniu (changes requested):
```
I’m concerned about the current coupling between the Agent and the Spreadsheet codepaths.

From the PR description it looks like the Agent is using existing spreadsheet APIs, but the change set appears to leak Agent-specific assumptions into the Spreadsheet area (or lets the Agent call deep spreadsheet internals). This increases coupling and maintenance surface area and makes it harder to test or replace either component independently.

Suggested fixes:
- Introduce a small adapter/interface (e.g. SpreadsheetDataProvider) that exposes only the minimal operations the Agent needs (readRange, getMetadata, writeCell, listSheets).
- Have the Spreadsheet module implement that interface and keep any UI/state concerns inside the Spreadsheet module.
- Make the Agent depend only on the interface (use DI or a factory) so we can mock it in tests.
- Add unit tests for the adapter contract and one integration test that exercises the Agent through the interface rather than by reaching into spreadsheet internals.
- Add a short developer doc describing the contract and intended responsibilities.
```
Why this matters
- Tighter separation reduces regressions, makes both components easier to test and evolve, and prevents accidental UI/UX or state coupling that leads to hard-to-debug bugs.


```
This integration gives the Agent access to spreadsheet content; I don’t see explicit handling for sensitive data, consent, or access boundaries.

Please add:
- A threat/PII checklist and enforce redaction or explicit opt-in for analyzing sheets that may contain personal or sensitive data.
- Permission checks: the Agent must only access ranges/sheets the user has explicitly allowed; surface an explicit consent step or feature flag for analysis.
- Limits on what is sent to any external services (if the Agent calls external LLMs): sanitize or redact PII before sending.
- Audit logging for analysis requests and a config switch to disable telemetry/collection for sensitive projects.
- Tests that validate redaction and permission behavior.
```
Why this matters
- Spreadsheets often contain sensitive personal or business data; insufficient controls risk data leaks, compliance issues, and user trust problems.

```
I’m worried about how the integration will behave on large spreadsheets (many rows/columns or large cell contents). The changes don’t appear to add protections for scale.

Please add:
- Defenses for large inputs (e.g. max cell/row/total-byte limits) and a graceful failure or fallback path (e.g. summarize first N rows, or ask user to narrow the range).
- Incremental/read-windowed loading instead of reading entire sheets into memory.
- Async/background processing for long-running analysis tasks and a way to show progress or cancel the job.
- Performance tests and a basic benchmark (small/medium/large sheet scenarios) to measure CPU/memory and latency.
- Timeouts and circuit-breakers when calling downstream LLMs or long-running processors.
```
Why this matters
- Without these limits and fallbacks, users with large spreadsheets can experience UI freezes, high memory/CPU usage, or timeouts — leading to poor UX and production incidents.

quanwangniuniu (approved):
Good job!

PR #796, opened 2026-09-01
Feature/med 327 enforce concurent login session limits

RainF0rester:
# Ticket Resolution Report — MED-327: Enforce Concurrent Login Session Limits

## 0. Summary

Introduces a Redis-backed `SessionRegistry` that tracks all active login sessions per user using a sorted set (score = login timestamp). On each login, the oldest sessions are automatically evicted when the user exceeds the org's `max_concurrent_sessions` cap.

Adds `max_concurrent_sessions` field to the `Organization` model (default: 5), configurable by Org Admins via a new Session Settings card on the organization settings page.

Exposes `GET /auth/sessions/` and `DELETE /auth/sessions/<jti>/` endpoints for listing and revoking active sessions. Session ownership is enforced — users can only revoke their own sessions.

Fixes logout to properly remove the session from Redis on sign-out using `request.auth.get("refresh_jti")` to avoid the `RefreshToken.blacklist()` incompatibility.

Adds session eviction UX: when a session is evicted (by cap overflow or explicit revoke), the next API request redirects immediately to `/login` with a toast notification, with no intermediate error flash.

## 1. Ticket Overview

| Field | Detail |
|-------|--------|
| Ticket ID | MED-327 |
| Title | Enforce Concurrent Login Session Limits |
| Assignee | yulin |
| Branch | feature/MED-327-enforce-concurent-login-session-limits |

## 2. Problem Description

**Background**

Before this ticket, there was no mechanism to limit the number of simultaneous active sessions per user. A user could log in from unlimited browsers, devices, or tabs with no visibility or control over their active sessions.

**Impact**

- No enforcement of concurrent session limits at the org level
- Users had no way to view or revoke their active sessions
- A compromised token could not be individually revoked without rotating the entire `auth_token_version`
- No UX for session eviction — evicted users would see a raw API error

**Related Modules**

- `authentication` — login, logout, token refresh, session registry
- `core` — Organization model, session cap setting
- frontend profile page, organization settings page

### New Module: `SessionRegistry` (`authentication/session_registry.py`)

Manages active sessions in Redis using three key patterns:

| Key Pattern | Type | Purpose |
|-------------|------|---------|
| `session:registry:{user_id}` | Sorted Set | All active JTIs, scored by login timestamp |
| `session:meta:{jti}` | Hash (via cache) | IP, user agent, login time per session |
| `session:blacklist:{jti}` | Key (via cache) | Marks evicted sessions |

TTL for all keys: 4 days (matches JWT refresh token lifetime).

**Methods:**

| Method | Description |
|--------|-------------|
| `register_session(user_id, jti, meta, cap)` | Adds session to sorted set; evicts oldest if over cap; returns list of evicted JTIs |
| `evict_session(user_id, jti)` | Blacklists JTI + removes from registry + deletes metadata |
| `remove_session(user_id, jti)` | Removes from registry without blacklisting (normal logout) |
| `is_evicted(jti)` | Returns True if JTI is on the blacklist |
| `list_sessions(user_id)` | Returns all active sessions with metadata |
| `delete_session(user_id, jti)` | Alias for `evict_session` (used by revoke endpoint) |

### New API Endpoints

**`GET /auth/sessions/`** — `SessionListView`

Returns all active sessions for the authenticated user. Each entry includes: `jti`, `ip`, `user_agent`, `created_at`.

**`DELETE /auth/sessions/<jti>/`** — `SessionRevokeView`

Revokes a specific session. Includes ownership check — verifies the JTI belongs to the current user before revoking. Returns 404 if not found.

**`POST /auth/token/refresh/`** — `SessionTokenRefreshView`

Wraps SimpleJWT's token refresh. The `refresh_jti` claim is embedded in the refresh token at login and auto-copied to every new access token by SimpleJWT, so session revocation remains effective after token refresh without manual injection.

### Modified: `LoginView`

After authenticating the user, registers the session in Redis and enforces the org cap:

```python
cap = user.current_organization.max_concurrent_sessions if user.current_organization else 5
evicted_jtis = SessionRegistry.register_session(user.pk, jti, meta, cap)
for evicted_jti in evicted_jtis:
    SessionRegistry.evict_session(user.pk, evicted_jti)
```

### Modified: `LogoutView`

On logout, removes the session from Redis using `refresh_jti` from the access token (avoids `RefreshToken.blacklist()` which is incompatible with `BLACKLIST_AFTER_ROTATION: True` without the token blacklist app installed):

```python
refresh_jti = request.auth.get("refresh_jti") if request.auth else None
if refresh_jti:
    SessionRegistry.remove_session(request.user.pk, refresh_jti)
```

### Modified: `TenantAwareJWTAuthentication` (`core/authentication.py`)

On every authenticated request, checks if the session's `refresh_jti` has been evicted:

```python
refresh_jti = validated_token.get("refresh_jti")
if refresh_jti and SessionRegistry.is_evicted(refresh_jti):
    raise AuthenticationFailed("Session has been evicted.", code="session_evicted")
```

### Modified: `Organization` model + `auth_tokens.py`

- Added `max_concurrent_sessions = PositiveBigIntegerField(default=5)` to `Organization`
- `build_user_refresh_token()` in `auth_tokens.py` embeds `refresh_jti` (= refresh token's own JTI) into the token payload, making it available in all derived access tokens

### Database Migration

`core/0023_organization_max_concurrent_sessions` — adds `max_concurrent_sessions` field to the `Organization` model.

### Frontend

**Profile page** (`frontend/src/app/(project)/profile/page.tsx`)

- New "Active Sessions" section listing all active sessions with browser/OS parsed from user agent, IP address, and login time
- Each session row has a "Revoke" button; revoking the current session redirects to login

**Organization Settings page** (`frontend/src/app/(project)/organizations/[orgId]/page.tsx`)

- New "Session Settings" card visible only to Org Admins
- Shows and allows editing the `max_concurrent_sessions` value

**Session eviction UX** (`frontend/src/lib/api.ts`, `frontend/src/app/(auth)/login/page.tsx`)

- API interceptor detects `session_evicted` 401 responses and calls `notifySessionEvicted()`, which clears all three auth storages (localStorage, sessionStorage, cookie) before redirecting to `/login`
- Login page reads `sessionStorage.getItem('session_evicted')` on mount and shows a toast: "Your session has been revoked. Please log in again."
- No intermediate error flash or redirect loop

**`AuthProvider`** (`frontend/src/components/providers/AuthProvider.tsx`)

- Removed dead `auth:session-evicted` event listener that previously caused a Chinese error message to flash before redirect

## 4. Testing

**Backend tests** (`authentication/tests/test_session_cap.py`)

| Test | What it verifies |
|------|-----------------|
| `test_register_session_under_cap` | Sessions under cap are all retained |
| `test_register_session_at_cap` | Sessions exactly at cap are retained |
| `test_register_session_over_cap_evicts_oldest` | Oldest session evicted when over cap |
| `test_evict_session_blacklists_jti` | Evicted JTI is blacklisted |
| `test_is_evicted_returns_true` | `is_evicted` returns True for blacklisted JTI |
| `test_is_evicted_returns_false` | `is_evicted` returns False for active JTI |
| `test_list_sessions_returns_metadata` | Session list includes ip, user_agent, created_at |
| `test_revoke_endpoint_returns_404_for_other_user` | Ownership check prevents cross-user revocation |

## 5. Lessons Learned

**`RefreshToken.blacklist()` incompatibility** — `BLACKLIST_AFTER_ROTATION: True` in SimpleJWT settings with `token_blacklist` app not installed causes `AttributeError` on any `RefreshToken` construction. The fix is to use `request.auth.get("refresh_jti")` in `LogoutView` — the access token already carries the `refresh_jti` claim, eliminating the need to touch the refresh token string.

**Daphne does not auto-reload** — Unlike Django's `runserver`, Daphne (used for ASGI/WebSocket support) does not detect file changes. Code changes require `docker restart backend-dev` to take effect.

**Three auth storages must be cleared on eviction** — The frontend persists auth state across localStorage, sessionStorage, and a cookie (`ms_auth`). Clearing only one or two caused a redirect loop: the page reloaded, re-hydrated from the remaining storage, saw `isAuthenticated = true`, and redirected back to `/overview`. `notifySessionEvicted()` must clear all three before redirecting.

**`remove_session` vs `evict_session`** — Normal logout should not blacklist the JTI (blacklisting is irreversible for the token's remaining lifetime). `remove_session` deletes the registry entry and metadata without adding to the blacklist, so a logged-out token cannot be replayed but also does not burn a blacklist slot.

## 6. References

- Jira Ticket: MED-327
- Branch: `feature/MED-327-enforce-concurent-login-session-limits`
- Key files:
  - `backend/authentication/session_registry.py`
  - `backend/authentication/views.py`
  - `backend/core/authentication.py`
  - `backend/core/services/auth_tokens.py`
  - `backend/core/migrations/0023_organization_max_concurrent_sessions.py`
  - `frontend/src/app/(project)/profile/page.tsx`
  - `frontend/src/app/(project)/organizations/[orgId]/page.tsx`
  - `frontend/src/lib/api.ts`
  - `frontend/src/app/(auth)/login/page.tsx`


quanwangniuniu (changes requested):
---

## Issue 1: **Race Condition in Session Eviction During Login**

### Problem

In `backend/authentication/views.py` (lines 319-325), when a user logs in and their session count exceeds the cap, the code registers the new session **first**, then manually evicts the oldest ones:

```python
evicted_jtis = SessionRegistry.register_session(user.pk, jti, meta, cap)
for evicted_jti in evicted_jtis:
    SessionRegistry.evict_session(user.pk, evicted_jti)
```

However, `SessionRegistry.register_session()` already handles eviction internally (line 32-34 in `session_registry.py`). This creates a **double-eviction problem**:
- The first eviction happens inside `register_session()` 
- Then you manually evict them again, which is redundant and can cause data inconsistency

Additionally, if `evict_session()` fails after the session is already registered, the user's session persists without being properly blacklisted.

### How to Fix

Remove the manual eviction loop. Let `SessionRegistry.register_session()` handle all eviction logic internally:

```python
# In LoginView.post() — lines 312-325
try:
    cap = user.current_organization.max_concurrent_sessions if user.current_organization else 5
    # register_session() already handles eviction internally
    SessionRegistry.register_session(user.pk, jti, meta, cap)
except Exception:
    logger.exception("Failed to register session in Redis for user %s", user.pk)
```

---

## Issue 2: **Missing Session Eviction on Concurrent Request**

### Problem

When a session is evicted (either by cap overflow or explicit revocation), there's no mechanism to immediately reject or clear that evicted session from the user's current connections. The eviction is only checked in `TenantAwareJWTAuthentication.get_user()` (line 49 in `core/authentication.py`), which means:

1. A user with an evicted token can continue making requests until the token naturally expires or they refresh
2. The eviction state is only checked on **next API call**, not proactively
3. No active websocket cleanup occurs (unlike logout, which has a websocket revoke in line 1186)

### How to Fix

Add proactive session invalidation to `SessionRegistry.evict_session()`:

```python
# In backend/authentication/session_registry.py
@staticmethod
def evict_session(user_id, jti):
    blacklist_key = BLACKLIST_KEY.format(jti=jti)
    register_key = REGISTER_KEY.format(user_id=user_id)
    redis = cast(Redis, get_redis_connection("default"))

    # Add to blacklist
    cache.set(blacklist_key, True, timeout=TOKEN_TTL)

    # Remove from registry
    redis.zrem(register_key, jti)

    # Delete metadata
    meta_key = META_KEY.format(jti=jti)
    cache.delete(meta_key)
    
    # NEW: Emit websocket revocation to active session
    try:
        from asgiref.sync import async_to_sync
        from channels.layers import get_channel_layer
        channel_layer = get_channel_layer()
        async_to_sync(channel_layer.group_send)(
            f'chat_user_{user_id}',
            {
                'type': 'user_session_revoked',
                'reason': 'session_evicted',
                'evicted_jti': jti,
            },
        )
    except Exception as e:
        logger.exception("Failed to emit session eviction websocket for user %s", user_id)
```

---

## Issue 3: **Metadata Storage Using Django Cache Without Explicit Expiry Handling**

### Problem

Session metadata (IP, user agent, created_at) is stored in Django's cache using `cache.set()` with a TTL:

```python
# Line 25 in session_registry.py
cache.set(meta_key, meta, timeout=TOKEN_TTL)
```

However, there are multiple issues:

1. **Inconsistent TTL management**: Redis sorted set keys use `redis.expire()` (line 24), but cache keys rely on Django's timeout parameter. If the cache backend changes or times out differently, metadata can disappear while the JTI still exists in the sorted set.

2. **No cleanup on eviction**: When `evict_session()` deletes metadata (line 51), if `cache.delete()` fails silently (e.g., key already expired), the sorted set and blacklist remain, causing orphaned entries.

3. **`list_sessions()` returns incomplete data**: Line 68-69 shows `meta = cache.get(meta_key) or {}`, which means if the cache expires, users see sessions with empty metadata but the JTI is still present.

### How to Fix

Implement explicit metadata cleanup and use a consistent Redis backend:

```python
# In backend/authentication/session_registry.py
import redis
from typing import cast

@staticmethod
def register_session(user_id, jti, meta: dict, cap: int) -> list[str]:
    redis_conn = cast(redis.Redis, get_redis_connection("default"))
    register_key = REGISTER_KEY.format(user_id=user_id)
    meta_key = META_KEY.format(jti=jti)

    # Add new session to sorted set, score = current timestamp
    redis_conn.zadd(register_key, {jti: time.time()})
    redis_conn.expire(register_key, TOKEN_TTL)

    # Store metadata directly in Redis Hash (not Django cache)
    redis_conn.hset(meta_key, mapping={
        'ip': meta.get('ip', ''),
        'user_agent': meta.get('user_agent', ''),
        'created_at': meta.get('created_at', ''),
    })
    redis_conn.expire(meta_key, TOKEN_TTL)

    # Check if over cap, evict oldest
    evicted = []
    excess = redis_conn.zcard(register_key) - cap
    if excess > 0:
        oldest = redis_conn.zpopmin(register_key, count=excess)
        evicted = [item[0].decode() if isinstance(item[0], bytes) else item[0] for item in oldest]
        # Clean up metadata for evicted sessions
        for evicted_jti in evicted:
            evicted_meta_key = META_KEY.format(jti=evicted_jti)
            redis_conn.delete(evicted_meta_key)

    return evicted

@staticmethod
def list_sessions(user_id) -> list[dict]:
    redis_conn = cast(redis.Redis, get_redis_connection("default"))
    register_key = REGISTER_KEY.format(user_id=user_id)

    jtis = redis_conn.zrange(register_key, 0, -1)

    sessions = []
    for jti_bytes in jtis:
        jti = jti_bytes.decode() if isinstance(jti_bytes, bytes) else jti_bytes
        meta_key = META_KEY.format(jti=jti)
        meta_dict = redis_conn.hgetall(meta_key)  # Returns dict with bytes keys/values
        meta = {
            'jti': jti,
            'ip': meta_dict.get(b'ip', b'').decode(),
            'user_agent': meta_dict.get(b'user_agent', b'').decode(),
            'created_at': meta_dict.get(b'created_at', b'').decode(),
        }
        sessions.append(meta)

    return sessions
```

---

quanwangniuniu (changes requested):
Scope: SessionRegistry (authentication/session_registry.py) — register_session concurrency/atomicity

The current register_session implementation can race under concurrent logins: if two (or more) logins for the same user run at the same time, both may observe the registry size < cap and skip evicting, allowing the number of active sessions to temporarily exceed the org's max_concurrent_sessions. This breaks the cap-enforcement guarantee.

Please make the add+trim operation atomic (so add the new JTI and evict oldest sessions in one atomic step). Two safe options:

1) Use a Redis Lua script (EVAL) that ZADDs the JTI, computes the overage, ZRANGEs the entries to evict, and performs the removals / blacklist writes / metadata deletes inside the same script; return the removed JTIs.
2) Or use Redis transactions with WATCH/MULTI/EXEC around the registry key, but prefer Lua for simpler and truly atomic semantics.


Why this matters
- Without atomic add+trim, concurrent logins can bypass the cap and temporarily create more active sessions than allowed, defeating the security/administrative intent of this feature.  
- This can cause inconsistent behavior (eviction UX not triggered, users seeing unexpected sessions), and makes audits or rate-based protections unreliable.  
- Implementing an atomic Redis operation (Lua or properly guarded transaction) ensures strong correctness under concurrency and avoids subtle, hard-to-reproduce bugs in production.

quanwangniuniu (approved):
Good job

PR #797, opened 2026-09-02
Feature/med 315 budget approval playwright e2e

FFFLAREEE:
## Summary
- Add `issue_budget_e2e_fixtures` management command and Playwright budget E2E specs (multi-step chain, single approver, reject/revise/resubmit, org-admin override).
- Fix V2 Task UI helpers, org-admin `make-approval` budget lookup, and double chain advance after admin override.
- Wire budget E2E into `application_test` CI with anchor provisioning, fixture generation, and Playwright trace/report artifacts on failure.
## Test plan
- [ ] `docker compose -p mediajira-v2 -f docker-compose.dev.yml exec backend pytest budget_approval/tests/test_issue_budget_e2e_fixtures.py budget_approval/tests/test_services.py -k override`
- [ ] Local E2E: generate fixtures then `cd frontend && E2E_USE_EXISTING_SERVER=1 BASE_URL=http://localhost npm run test:e2e:budget`
- [ ] CI `application_test` passes (budget Playwright step + artifact upload)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,7 @@ backend/celerybeat.pid
 # run_timestamp every test run; nothing reads it as a baseline).
 backend/benchmarks/retrospective.json
 test-results/chat-load/
+frontend/e2e/budget/fixtures/budget-e2e-fixtures.json
 
 # Debug/scratch scripts — never tracked
 frontend/.scratch/
--- a/.github/workflows/mediajira-ci.yml
+++ b/.github/workflows/mediajira-ci.yml
@@ -555,6 +555,139 @@ jobs:
               npm test
             "
 
+      # Budget approval Playwright E2E (MED-315): real backend via nginx, isolated fixtures per test.
+      - name: Wait for nginx (budget E2E)
+        run: |
+          echo "Waiting for nginx to serve the stack..."
+          timeout=120
+          counter=0
+          while [ $counter -lt $timeout ]; do
+            if curl -fsS http://localhost/health > /dev/null 2>&1; then
+              echo "nginx is ready!"
+              exit 0
+            fi
+            echo "nginx not ready yet, waiting... ($counter/$timeout)"
+            sleep 2
+            counter=$((counter + 2))
+          done
+          echo "nginx failed to become healthy within ${timeout}s"
+          docker compose logs nginx
+          exit 1
+
+      - name: Provision budget E2E anchor user
+        run: |
+          docker compose exec -T backend sh <<'EOS'
+          set -e
+          python manage.py shell <<'PY'
+          from django.contrib.auth import get_user_model
+          from django.db.models.signals import post_save
+
+          from core.models import Organization, OrganizationMembership, Project, ProjectMember
+          from core.services.tenant import provision_tenant_schema, slug_to_schema_name
+          from core.tenant_context import tenant_schema_context
+          from stripe_meta.signals import create_default_subscription
+
+          User = get_user_model()
+          ANCHOR_EMAIL = "ci-budget-anchor@example.com"
+          ORG_SLUG = "ci-budget-e2e-org"
+
+          post_save.disconnect(create_default_subscription, sender=Organization)
+          try:
+              org, _ = Organization.objects.get_or_create(
+                  slug=ORG_SLUG,
+                  defaults={"name": "CI Budget E2E Org", "is_active": True},
+              )
+              provision_tenant_schema(org.slug)
+
+              anchor, _ = User.objects.get_or_create(
+                  email=ANCHOR_EMAIL,
+                  defaults={
+                      "username": "ci-budget-anchor",
+                      "organization": org,
+                      "current_organization": org,
+                      "is_verified": True,
+                      "password_set": True,
+                  },
+              )
+              anchor.organization = org
+              anchor.current_organization = org
+              anchor.is_verified = True
+              anchor.password_set = True
+              anchor.set_password("password123!")
+              anchor.save()
+
+              OrganizationMembership.objects.update_or_create(
+                  user=anchor,
+                  organization=org,
+                  defaults={"role": "admin", "is_active": True},
+              )
+
+              schema = slug_to_schema_name(org.slug)
+              with tenant_schema_context(schema):
+                  project, _ = Project.objects.get_or_create(
+                      organization=org,
+                      name="CI Budget Anchor Project",
+                      defaults={"owner": anchor},
+                  )
+                  ProjectMember.objects.update_or_create(
+                      user=anchor,
+                      project=project,
+                      defaults={"role": "owner", "is_active": True},
+                  )
+              anchor.active_project = project
+              anchor.save(update_fields=["active_project"])
+              print("budget E2E anchor ready:", anchor.email, "project_id=", project.id)
+          finally:
+              post_save.connect(create_default_subscription, sender=Organization)
+          PY
+          EOS
+
+      - name: Generate budget E2E fixtures
+        run: |
+          set -e
+          mkdir -p frontend/e2e/budget/fixtures
+          docker compose exec -T backend python manage.py issue_budget_e2e_fixtures \
+            --allow-outside-debug \
+            --anchor-email=ci-budget-anchor@example.com \
+            > frontend/e2e/budget/fixtures/budget-e2e-fixtures.json
+          python3 -c "
+          import json
+          data = json.load(open('frontend/e2e/budget/fixtures/budget-e2e-fixtures.json'))
+          assert data.get('users', {}).get('requester', {}).get('access_token'), 'missing requester access_token'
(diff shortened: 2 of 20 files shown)


quanwangniuniu (changes requested):
---

## **Issue 1: Missing `ai_analysis_enabled` Column in Database Schema Migration**

### **Where:**
- `variations-studio-api/tests/support/fixtures.ts` (line 133)
- `variations-studio-api/tests/support/tenantRoutingFixtures.ts` (line 126)

### **The Problem:**
The test fixtures are inserting `ai_analysis_enabled: true` into the `core_project` table, but there's **no database migration** in this PR to add this column. This will cause:
- `IntegrityError` when test fixtures try to insert into the column
- Tests fail with: `column "ai_analysis_enabled" of relation "core_project" does not exist`
- The CI `application_test` job will fail at fixture setup time

### **How to Fix:**
Create a Django migration in `backend/core/migrations/`:
```python
# 000N_add_ai_analysis_enabled.py
from django.db import migrations, models

class Migration(migrations.Migration):
    dependencies = [
        ('core', 'XXXX_previous_migration'),
    ]

    operations = [
        migrations.AddField(
            model_name='project',
            name='ai_analysis_enabled',
            field=models.BooleanField(default=False),
        ),
    ]
```

Then update `backend/core/models.py` to add the field to the `Project` model.

### **Why This Matters:**
Without the schema migration, the fixture setup will silently fail during CI, making the E2E tests unreliable. The PR runs `issue_budget_e2e_fixtures` which depends on the anchor project being created—if schema migrations don't match test expectations, the entire test suite bootstrapping fails.

---

## **Issue 2: Unsafe Assumption About Budget Request Lookup in `task/views.py`**

### **Where:**
`backend/task/views.py` line 1382-1387 (the `make_approval` method)

### **The Problem:**
```python
br = task.linked_object or task.budget_requests.first()
if br is None:
    br = (
        BudgetRequest.objects.filter(task_id=task.pk)
        .order_by('-id')
        .first()
    )
```

The fallback logic queries `BudgetRequest.objects.filter()` **without tenant context**. In a multi-tenant system:
- If a task spans multiple tenants, this query **escapes the current tenant schema**
- It may return a budget request from a **different organization**
- This creates a **cross-tenant data leak** where one org's approver can approve another org's budget

This is particularly risky because `make_approval` is called after the org-admin override fix that already deals with approval chain logic—introducing the wrong budget request here silently corrupts the approval flow.

### **How to Fix:**
Wrap the fallback query in tenant context:
```python
from core.tenant_context import tenant_schema_context
from core.services.tenant import slug_to_schema_name

# At the start of make_approval, ensure tenant schema:
if task.project:
    schema = slug_to_schema_name(task.project.organization.slug)
    with tenant_schema_context(schema):
        br = task.linked_object or task.budget_requests.first()
        if br is None:
            br = (
                BudgetRequest.objects.filter(task_id=task.pk)
                .order_by('-id')
                .first()
            )
```

Or better—add a `.select_related('task__project__organization')` and enforce schema isolation at the model manager level.

### **Why This Matters:**
The org-admin override feature in this PR is a critical security-sensitive flow. A cross-tenant lookup bug means:
1. **Authorization bypass**: Org A's admin could inadvertently approve Org B's budgets
2. **Audit trail corruption**: Approval records get attached to the wrong task
3. **Approval chain desynchronization**: The test `test_org_admin_override_does_not_double_advance_after_task_sync` may pass locally (single-tenant) but fail in multi-tenant CI if tenant isolation isn't enforced

---

FFFLAREEE:
> ## **Issue 1: Missing `ai_analysis_enabled` Column in Database Schema Migration**
> ### **Where:**
> * `variations-studio-api/tests/support/fixtures.ts` (line 133)
> * `variations-studio-api/tests/support/tenantRoutingFixtures.ts` (line 126)
> 
> ### **The Problem:**
> The test fixtures are inserting `ai_analysis_enabled: true` into the `core_project` table, but there's **no database migration** in this PR to add this column. This will cause:
> 
> * `IntegrityError` when test fixtures try to insert into the column
> * Tests fail with: `column "ai_analysis_enabled" of relation "core_project" does not exist`
> * The CI `application_test` job will fail at fixture setup time
> 
> ### **How to Fix:**
> Create a Django migration in `backend/core/migrations/`:
> 
> ```python
> # 000N_add_ai_analysis_enabled.py
> from django.db import migrations, models
> 
> class Migration(migrations.Migration):
>     dependencies = [
>         ('core', 'XXXX_previous_migration'),
>     ]
> 
>     operations = [
>         migrations.AddField(
>             model_name='project',
>             name='ai_analysis_enabled',
>             field=models.BooleanField(default=False),
>         ),
>     ]
> ```
> 
> Then update `backend/core/models.py` to add the field to the `Project` model.
> 
> ### **Why This Matters:**
> Without the schema migration, the fixture setup will silently fail during CI, making the E2E tests unreliable. The PR runs `issue_budget_e2e_fixtures` which depends on the anchor project being created—if schema migrations don't match test expectations, the entire test suite bootstrapping fails.
> 
> ## **Issue 2: Unsafe Assumption About Budget Request Lookup in `task/views.py`**
> ### **Where:**
> `backend/task/views.py` line 1382-1387 (the `make_approval` method)
> 
> ### **The Problem:**
> ```python
> br = task.linked_object or task.budget_requests.first()
> if br is None:
>     br = (
>         BudgetRequest.objects.filter(task_id=task.pk)
>         .order_by('-id')
>         .first()
>     )
> ```
> 
> The fallback logic queries `BudgetRequest.objects.filter()` **without tenant context**. In a multi-tenant system:
> 
> * If a task spans multiple tenants, this query **escapes the current tenant schema**
> * It may return a budget request from a **different organization**
> * This creates a **cross-tenant data leak** where one org's approver can approve another org's budget
> 
> This is particularly risky because `make_approval` is called after the org-admin override fix that already deals with approval chain logic—introducing the wrong budget request here silently corrupts the approval flow.
> 
> ### **How to Fix:**
> Wrap the fallback query in tenant context:
> 
> ```python
> from core.tenant_context import tenant_schema_context
> from core.services.tenant import slug_to_schema_name
> 
> # At the start of make_approval, ensure tenant schema:
> if task.project:
>     schema = slug_to_schema_name(task.project.organization.slug)
>     with tenant_schema_context(schema):
>         br = task.linked_object or task.budget_requests.first()
>         if br is None:
>             br = (
>                 BudgetRequest.objects.filter(task_id=task.pk)
>                 .order_by('-id')
>                 .first()
>             )
> ```
> 
> Or better—add a `.select_related('task__project__organization')` and enforce schema isolation at the model manager level.
> 
> ### **Why This Matters:**
> The org-admin override feature in this PR is a critical security-sensitive flow. A cross-tenant lookup bug means:
> 
> 1. **Authorization bypass**: Org A's admin could inadvertently approve Org B's budgets
> 2. **Audit trail corruption**: Approval records get attached to the wrong task
> 3. **Approval chain desynchronization**: The test `test_org_admin_override_does_not_double_advance_after_task_sync` may pass locally (single-tenant) but fail in multi-tenant CI if tenant isolation isn't enforced

Thanks for raising these points. I checked both against the current `prod-preview` implementation.

**Issue 1 — `ai_analysis_enabled` migration**

`ai_analysis_enabled` is not introduced by this PR. The field and its database migration already exist in `prod-preview` as `core` migration `0023`, from the pivot/spreadsheet work.

This PR only updates the Variations Studio Jest fixtures so their raw `core_project` inserts include the existing non-null column and no longer fail with PostgreSQL error `23502`.

Adding another `AddField` migration here would duplicate and conflict with the existing `0023` migration. The suggested `default=False` would also be inconsistent with the current `Project` model, where the field defaults to `True`.

Therefore, no additional migration is required in this PR.

**Issue 2 — `BudgetRequest` tenant isolation**

`BudgetRequest` is tenant-scoped. For the `make_approval` HTTP path, `TenantSchemaMiddleware` has already set the PostgreSQL `search_path` to the active organization schema before the view executes.

As a result:

```python
BudgetRequest.objects.filter(task_id=task.pk)
```

queries only the `BudgetRequest` table in the current organization schema; it does not scan or read tables from other tenant schemas.

The fallback is used only when the task’s GenericFK/`linked_object` is missing, allowing the view to resolve the related `BudgetRequest` through `task_id` within the same tenant schema. This fixes the org-admin override path incorrectly returning 403.

Wrapping the query in another `tenant_schema_context` would duplicate the request-level middleware behavior. It would also not replace or correct tenant resolution at the request boundary if that resolution were wrong.

.gitignore
```diff
@@ -39,6 +39,7 @@ backend/celerybeat.pid
 # run_timestamp every test run; nothing reads it as a baseline).
 backend/benchmarks/retrospective.json
 test-results/chat-load/
+frontend/e2e/budget/fixtures/budget-e2e-fixtures.json
```
quanwangniuniu, line 42: does this file even exist?

quanwangniuniu (approved):
Good job

PR #799, opened 2026-09-02
Implement claude dev convention structure

07Coder29:
Set up Claude Code project structure (CLAUDE.md, rules, agents, skills, hooks)

### DIFF
--- a/.claude/skills/deploy/SKILL.md
+++ b/.claude/skills/deploy/SKILL.md
@@ -1,0 +1,44 @@
+---
+name: deploy
+description: How MediaJira ships to production, and where a developer's responsibility ends. Use when asked to deploy, release, ship, promote to main, cut a release, or roll back.
+disable-model-invocation: true
+---
+
+# Deploying MediaJira
+
+## A developer's responsibility ends at `prod-preview`
+
+The only branch flow a developer (or Claude) drives is **feature branch → reviewed PR → merged
+into `prod-preview`**. That's it.
+
+**The `prod-preview` → `main` promotion is not a developer task.** It is owned by whoever runs
+releases. Claude must **not**:
+
+- open, draft, or prepare a `prod-preview` → `main` PR,
+- merge to `main`, push to `main`, or trigger the deploy webhook,
+- run `recompose.sh` or any production compose command.
+
+When someone asks Claude to "deploy" or "release", the answer is: make sure the change is
+reviewed and merged into `prod-preview`, then hand off to the release owner. Everything below is
+**context only** — so Claude can answer questions, not so it can act.
+
+## What happens after `prod-preview` (for reference)
+
+1. The release owner opens `prod-preview` → `main` (the only branch pair CI's `block` job lets
+   PR into `main`).
+2. Merging to `main` runs `.github/workflows/mediajira-ci.yml`:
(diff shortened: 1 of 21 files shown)


quanwangniuniu (changes requested):
1. **The Code Reviewer Agent is not implemented correctly** and needs to be revisited. Also, please note that our architecture includes both **Node.js and .NET microservices**. The .NET microservice is currently coming — please contact **Gigi** for the latest details and integration requirements.

2. **The Hooks implementation needs significantly more validation and edge-case coverage.** The current implementation does not yet meet an industry-level standard. Please consider additional validation points, failure scenarios, boundary conditions, concurrency issues, invalid inputs, and other relevant edge cases to make the implementation robust and production-ready.

3. **`CLAUDE.md` currently exists only at the root module level**, which is not ideal from a modularity and maintainability perspective. This also appears to be inconsistent with the recommended usage patterns described in the **official Claude Code documentation**, where project/module-specific instructions can be scoped closer to the relevant code. We should restructure the `CLAUDE.md` files so that instructions are appropriately scoped to each module/layer, improving modularity and ensuring Claude Code receives the right context when working within a specific part of the codebase.

.claude/hooks/check-edit.sh
quanwangniuniu: lets remove all the .sh or similar script since they wont work on mac

.claude/hooks/tests/test_check_conventions.py
quanwangniuniu: This one wont work. Did you even run this? I dont think we even need this and it is too heavy.

.claude/skills/deploy/SKILL.md
quanwangniuniu: this project is not mediajia

plus, the deployment method is totally AI generated and not true

docs/ARCHITECTURE.md
quanwangniuniu: I dont understand why this one is here, AI would never read it. And even AI read this one, it lose focus and it wont help anything

scripts/conventions/check_conventions.py
quanwangniuniu: I dont think this one could work

.mcp.json.example
quanwangniuniu: Why we need this?

env.example
quanwangniuniu: This one is out of scope

CLAUDE.md
quanwangniuniu: Too many scripts in this file, which result in losing the purpose of this whole file

.mcp.json
quanwangniuniu: what is the MCP database url and do we even have this one?

CLAUDE.local.md.example
quanwangniuniu: Since you mentioned personal ones should not be commited, why this one is here? It does not make any sense

.github/workflows/mediajira-ci.yml
quanwangniuniu: this one might break the ci workflow on github actions so revert this

quanwangniuniu (changes requested):
11 comments need to be dealt with. Also I dont believe all the files in rules could be used. As I mentioned before, this one should have index style + modularity harness for AI to refer. But current change still looks like we are trying to dumping 100 files to AI and try to let it listen

PR #801, opened 2026-09-04
fix(security): drive CORS/CSRF extra origins from env vars (MED-398)

wenxiaogao000:
Remove hardcoded ngrok tunnel hostnames from ALLOWED_HOSTS, CORS_ALLOWED_ORIGINS, and CSRF_TRUSTED_ORIGINS. Free-tier tunnel domains are not permanently owned and could be reclaimed by an unrelated party who would then be trusted for credentialed cross-origin requests. Dev-only extra origins now come from CORS_EXTRA_ORIGINS / CSRF_EXTRA_ORIGINS env vars, left empty in production.


quanwangniuniu (approved):
Good job

PR #802, opened 2026-09-06
Med 317 raise agent module test coverage to 60

leo-maxwell:
1. fixed the startup shell problem
2. finished the ticket MED-317 on linear
3. fixed three problems:
1) wrong llm_call function is used in executors.py and services.py
2) auto-categorization problem in column_registry.py
3) json serialization problem in services.py

the new test coverage is 63% and currently all tests pass.


leo-maxwell:
The original commit did not consider the fact that in the latest prod-preview branch the entry point of gemini call function has changed. The second commit have fixed this issue.

docker-compose.dev.yml
quanwangniuniu, line 1: This file is not relevent

docker-compose.dev.yml
```diff
@@ -61,7 +61,7 @@ services:
       - "host.docker.internal:host-gateway"
       - "postgres-host:host-gateway"
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -h 127.0.0.1 -p 5432 -U ${POSTGRES_USER:-postgres} -d ${POSTGRES_DB:-mediajira_db}"]
+      test: ["CMD", "pg_isready", "-h", "127.0.0.1", "-p", "5432", "-U", "${POSTGRES_USER:-postgres}", "-d", "${POSTGRES_DB:-mediajira_db}"]
```
quanwangniuniu, line 64: this file is not relevent

quanwangniuniu (changes requested):
This ticket should only involve test files, all other changes are not allowed. or if you change them, can you tell me the reasons?

docker-compose.dev.yml
leo-maxwell: It is indeed not relevant but I would take this as a genuine improvement to prevent name expansion issues, as the original command would fail on my setup. This change does not affect its behaviour at all, the deployment should go like before on everyone else's computer. If it is not seen as an issue, please ignore it.

quanwangniuniu: cool

quanwangniuniu (approved):
Good job

PR #805, opened 2026-09-08
Med 128 e2e playwright tests ads pages

leo-maxwell:
fixed a few backend bugs and added tests for them while writing tests for the frontend.


docker-compose.e2e.yml
quanwangniuniu, line 1: why we need this one?

backend/authentication/tests/test_delete_account.py
quanwangniuniu, line 1: why we have test_delete_account.py?

backend/core/services/tenant.py
quanwangniuniu, line 1: why we change this one?

backend/core/models.py
quanwangniuniu, line 1: why?

backend/google_ads/views.py
quanwangniuniu, line 1: why?

frontend/package.json
quanwangniuniu, line 1: why we change versions for dependency

quanwangniuniu (changes requested):
seems we have changes unrelated.

docker-compose.e2e.yml
leo-maxwell: This file was added because we need an environment for a test: actual picture preview. The test itself uploads an image, generates a public preview link, and verifies if the picture is correctly loaded.

Currently the public preview page of the Facebook Ads resolves relative media URLs against `NEXT_PUBLIC_API_URL`, falling back to http://localhost:8000 when that variable is not set. Django’s development media routes are only registered when `DEBUG=True`. Hence we need this override to enable the Django local media service, and publishes the port 8000 to the local loopback so that the testing browser is able to access.

It currently only works when explicitly included in the `docker compose` command. Adding this file does not change the normal deployment configuration.

There is currently an integration gap: no workflow or script is referring to this file. I think we should keep the image-loading assertion because a public preview with a broken image is a user-visible failure that checking the page or API response alone would miss. We should wire this override into the relevant E2E test command, or use an existing test configuration that provides equivalent media access. Can be removed without sacrificing the image loading test if another configuration already meets that requirement.

backend/authentication/tests/test_delete_account.py
leo-maxwell: The E2E tests create real accounts and delete them during cleanup. This regression test ensures account deletion still succeeds when the user’s active project no longer exists

backend/core/services/tenant.py
leo-maxwell: During local E2E testing concurrent workers creating their test organizations triggered a PostgreSQL deadlock due to conflicting locks on shared public tables. This caused organization creation to return HTTP 500.

The old error-handling code then tried to reset `search_path` inside the failed transaction. That SQL statement also failed, masking the original error.

This change serializes tenant initialization with a transaction-level advisory lock and keeps table creation and foreign-key setup within a transaction. On success, it restores the caller’s `search_path`; on failure, rollback restores it without running cleanup SQL inside the failed transaction. Regression tests covering five concurrent organization creations, foreign-key creation failures, and connection recovery all pass.

This also affects production: concurrent organization creations must wait for the transaction holding the lock to finish. That is the trade-off for preventing the deadlock reproduced during testing.

backend/core/models.py
leo-maxwell: This modification is to adapt to the problem fixed in `tenant.py`. The organization creations workflow is adjusted this way:

1. Obtain the lock
2. Initialize the tenant structure
3. Submit the transaction

This logic is only used when creating a new organization.

backend/google_ads/views.py
leo-maxwell: These changes fix three bugs found while developing the E2E tests:

1. Incorrect creator assignment. The view passed `created_by=user`, but the custom serializer expected `created_by_id` and defaulted it to `None`. Passing both values to the model cleared the creator association. The view now passes the authenticated user’s ID in the format the serializer expects.
2. Owners could not access drafts without a customer account. The queries included ads whose customer account was created by the current user, or ads with no creator, but missed ads created directly by the current user. Once creator assignment was fixed, accountless drafts were excluded. Adding the creator condition lets owners list, open, update, and delete these drafts.
3. The generated share URL did not match the frontend route. The API returned `/google_ads/preview/<token>/`, but the existing page expects `/google_ads/preview/share?share=…`, with an encoded payload containing `previewToken` and `device`. The returned URL now matches that page’s expected format.

frontend/package.json
leo-maxwell: The playwright version 1.58.2 has problem when tidying and merging trace zip files. The bug is described in https://github.com/microsoft/playwright/pull/40747 and version 1.60.0 fixes it.

quanwangniuniu (approved):
Good job

PR #800, opened 2026-09-03
test(MED-314): raise Jest coverage threshold to 30% for lib/ui scope

FFFLAREEE:
## Summary
- Scope Jest coverage collection to `src/lib/**` and `src/components/ui/**`, and set the global floor to 30%.
- Remove overly strict per-file thresholds (miro / PermissionMatrix / ApproverSelect 80–85%).
- Backfill unit tests for lib helpers, API clients, and UI components; CI now runs `npm run test:ci` so PRs fail below the threshold.
## Test plan
- [ ] `cd frontend && npm run test:ci` passes locally (coverage ≥ 30% in scope)
- [ ] CI `application_test` frontend step is green
- [ ] Storybook / Chromatic a11y remains green
- [ ] Confirm unrelated files (`variations-studio-api`, `e2e/budget/fixtures`) were not included

PR #806, opened 2026-09-09
feat(calendar): Calendly-style booking links + MED-284

zizhenliu0427:
## Summary

Completes MED-284 — Calendly-style booking links on Calendar. A host publishes a shareable link; an external prospect chooses an available time and creates an in-app meeting without an account. Previously that meeting had to be arranged through back-and-forth messages. Links are managed from **Create Booking link**, to the left of Ask Agent on the Calendar toolbar.

- **Link management** — create, edit, disable and delete links with duration, slot increments, buffers, minimum notice, advance booking limits, timezone and weekly availability windows. Team links write the host event to the selected project calendar; Personal links write it to the selected personal calendar. Default type is Personal.
- **One host and one booker** — the invitee list names people who may each book the link; it does not add everyone to the same meeting. Project colleagues, project CSM contacts and guest email addresses can be selected. Existing owner/admin permissions for creating a colleague's link remain in place.
- **Public booking page** — `/book/<org>/<slug>` with APIs under `/api/public/book/<org>/<slug>/`. The organisation slug selects the tenant schema. Unknown organisations and links return the same not-found response. Restricted links require an authorised invitee.
- **Identity** — anonymous bookers supply a name and email, with an optional telephone number. Signed-in bookers use their account identity. Entering a registered email anonymously does not grant that account or write to its personal calendar.
- **Availability** — slots respect weekly windows, duration, notice, buffers, recurring events and timezones. In-app busy time currently comes from the calendar selected by the link. Personal links can also consult the host's Google connection when one exists; Team links do not.
- **Confirmation and copies** — a booking creates the host event and, for an authenticated booker in the same organisation, a personal-calendar copy. Related copies stay in step on cancel. A returning signed-in booker is reminded of their existing date and time on the picker.
- **Cancellation, recovery and subscriptions** — the host can delete the calendar event; the guest can cancel from the tokenised page or recover the booking by email. Confirmation email is queued with management and subscription links. The confirmation page offers a webcal/HTTP feed and a downloadable ICS snapshot.
- **Reliability** — booking submission rechecks availability inside a transaction with calendar locks. Buffer-aware and revoke-safe overlap checks cover competing bookings on the same calendar. Recovery does not disclose cancellation tokens to someone who only knows a name, email or telephone number.
- **CI unblock after merging `prod-preview`** — not part of the booking ticket. The shared pipeline now runs budget Playwright after `npm run build` inside the live frontend container. That build replaces `.next` while Next.js is still running, so `/_next/static/*.js` 404s and the suite fails even when nginx is healthy. This PR restarts the frontend and checks that the login page and its scripts actually load before Playwright starts. Without that, this branch cannot pass CI.

## Files Changed

(121 files; key areas)

**CI**
- `.github/workflows/mediajira-ci.yml` — restart frontend after the in-container build and verify login-page assets before budget Playwright

**Backend — model, public API, availability**
- `backend/calendars/models.py` — `BookingLink`, invitees, attendee phone and booking metadata
- `backend/calendars/serializers.py` / `views.py` / `urls.py` — owner CRUD plus public availability, create, cancel, lookup and ICS feed
- `backend/calendars/availability.py` / `services.py` — slot calculation and merged busy intervals
- `backend/calendars/migrations/0005_*.py` through `0011_*.py` — booking-link schema
- `backend/core/tenant_config.py` / `backend/core/services/tenant.py` — tenant schema synchronisation for new columns

**Backend — write path, recovery, notifications, tests**
- `backend/calendars/booking_write.py` — transactional event creation, related copies and cancellation
- `backend/calendars/booking_access.py` / `booking_lookup.py` / `booking_tokens.py` / `booking_ics.py` / `booking_notifications.py` / `booking_invite_state.py` / `tasks.py` — access, recovery, subscriptions and mail
- `backend/google_calendar_integration/services.py` / `tasks.py` — existing free/busy and export helpers; live Google write remains deferred
- `backend/calendars/test_public_booking.py`, `test_booking_*.py`, `test_availability.py` and Google integration tests — public contract, concurrency, recovery, ICS and visibility

**Frontend — public booking**
- `frontend/src/app/(public)/book/[orgSlug]/[linkSlug]/page.tsx` / `cancel/page.tsx` — public picker and cancel routes
- `frontend/src/components/calendar/BookingWidget.tsx` — date/time picker, confirm form, existing-booking reminder and confirmation
- `frontend/src/components/calendar/BookingCancel.tsx` / `BookingFind.tsx` / `BookingExistingReminder.tsx` — cancel, find-again and returning-booker reminder
- `frontend/src/lib/api/calendarApi.ts` — public and owner booking clients

**Frontend — link management, calendar, notifications**
- `frontend/src/components/calendar/BookingLinkManager.tsx` / `CalendarToolbar.tsx` — Create Booking link and Team/Personal setup
- `frontend/src/components/calendar/CalendarPageContent.tsx` / `CalendarSidebarPanel.tsx` / `EventPanelDialog.tsx` — My Calendar / project filtering and booking-event actions
- `frontend/src/components/notifications/drawer/*` — booking invite lifecycle in the notification drawer
- `frontend/e2e/calendar/booking-link.spec.ts` / `booking-link-manager.spec.ts` / `booking-link-real.spec.ts` — mocked and real-backend Playwright coverage

## Expected Result

- A host can create a Personal or Team booking link from Calendar, copy `/book/<org>/<slug>`, and have slots match the published rules with busy times excluded.
- An anonymous prospect can book with name and email (telephone optional) and receive a confirmation screen with cancel and subscribe options.
- A signed-in project invitee can book without re-typing identity; a same-project booker is guided to Personal when Team is not allowed on that link.
- The meeting appears on the host's selected calendar, and a same-organisation signed-in booker also sees a personal-calendar copy.
- Either side can cancel: host delete or guest cancel/find-again. Related copies are cancelled together and the slot is released.
- A returning signed-in booker sees which day and time they already booked when they reopen the picker.
- Unknown or other-organisation links look the same as a missing link and do not leak owner email or internal identifiers.
- Double-booking the same slot is rejected rather than silently succeeding.
- Confirmation and recovery mail is queued with meeting and management links. Inbox delivery was not verified: there is no API to inspect the outbox or prove a message reached a mailbox.

## Follow-up Recommendation

Live Google OAuth and Calendar write were parked at the ticket owner's request; existing free/busy and export helpers stay in the Google package for a later connection. Ordinary project members creating links for other people, multiple required participants in one meeting, and team round-robin remain outside this ticket.

A few behaviours should be reconciled on a follow-up rather than implied to be finished here: the public Team/Personal selector is not submitted to the API, so the confirmation copy can disagree with the actual destination; availability is checked against the selected calendar rather than every diary belonging to the host; booking another time replaces the earliest upcoming linked booking for a signed-in user but creates an additional booking for an anonymous visitor; replacement bookings receive a new event ID, so an existing feed does not automatically follow the new meeting.

Existing-tenant upgrades need attention. Calendar migrations now run through `0011`. A clean `makemigrations --check` does not prove tenant schemas are ready. After merging `prod-preview`, full `migrate_tenant` on the local QA tenant was blocked by the new non-null chat `seq` field, which needs its data backfill replayed; only the Project schema used for booking tests was repaired.

Confirmation / recovery email delivery was not tested. There is no API to inspect queued mail or prove inbox receipt; coverage stops at enqueueing the send task. Proving delivery needs a worker that has loaded the new task, a working mail service, and a mailbox check outside this PR. Cloud calendar clients cannot fetch a developer's localhost feed URL.

The frontend-restart CI step should move onto `prod-preview` itself. It is a pipeline defect, not a booking-link change; other branches that have already picked up the budget Playwright job will keep hitting the same asset mismatch until the workflow fix is on the base branch.

## Test Plan

- [x] Backend impacted tests: `pytest -q -n 0 --no-cov calendars google_calendar_integration` — **489 passed**
- [x] Frontend Jest (booking/calendar suites): **19 suites / 109 tests passed**
- [x] Playwright page scenarios: `npx playwright test e2e/calendar/booking-link.spec.ts e2e/calendar/booking-link-manager.spec.ts` — **31 passed**, plus login setup
- [x] Playwright real-backend scenarios: `npx playwright test e2e/calendar/booking-link-real.spec.ts` — **5 passed**, plus login setup (anonymous booking, duplicate-slot rejection, one successful concurrent submission, before-buffer enforcement, cancellation, feed updates and slot release)
- [x] `makemigrations --check --dry-run` — no changes detected
- [x] Frontend lint — no errors; six warnings, including a missing `refetch` effect dependency in `CalendarPageContent`
- [x] Manual QA passed; demonstration video: https://drive.google.com/file/d/1yDmBghTUF6-4jjM-vfmo4kX9HXbSHdoU/view?usp=sharing
- [x] PR CI (`application_test`) passed after the frontend restart / asset check; the previous run failed at budget Playwright while nginx was already healthy

Results were recorded against merge commit `b9c234e04`, incorporating `prod-preview` at `7837810d3`. Google integration tests use mocks. These results do not claim live Google writes or native calendar-client refresh.

Ticket: [MED-284](https://linear.app/mediajira/issue/MED-284/calendar-calendly-style-booking-links)


quanwangniuniu (changes requested):
7 Critical Problems with This PR

Based on my analysis of PR #806 (Calendly-style booking links + MED-284), here are 7 significant problems that should be addressed:

### 1. **Massive Scope (15,556 additions, 352 deletions) - High Risk of Regressions**
**Why fix it:** This PR introduces an enormous feature spanning 121 files with complex calendar availability logic, multi-tenancy context switching, and public API endpoints. Changes of this magnitude carry elevated risk of:
- Subtle interactions with existing calendar features
- Performance degradation on edge cases
- Hard-to-spot security vulnerabilities in the new public endpoints
- Testing gaps that won't surface until production

**Recommendation:** Consider splitting this into smaller, independently reviewable PRs (model + migrations, availability logic, public endpoints, notifications, etc.)

---

### 2. **CI Restart Heuristic is Fragile (.github/workflows/mediajira-ci.yml)**
**Why fix it:** The PR modifies CI to restart the frontend and check that login page scripts load. The problem:
- The heuristic (checking if `/_next/static/*.js` returns 200) is brittle—a partial failure or network hiccup in that window breaks the suite unnecessarily
- No exponential backoff; the 2-second sleep is hardcoded, risking race conditions
- The Python script in the shell runs over 70 lines inline; it's unmaintained and hard to debug if it fails

**Recommendation:** Extract the readiness check into a separate, testable script with proper retry logic and clearer error messages.

---

### 3. **Multi-Tenant Schema Switching Without Explicit Rollback Protection**
**Why fix it:** Public booking endpoints manually call `tenant_schema_context()` to switch schemas (e.g., in `PublicBookingCreateView`). If an exception occurs:
- The `with` block exits and context is restored, BUT
- Uncommitted transaction state is NOT guaranteed to be rolled back in the worker queue tasks
- Tasks queued inside a failed atomic block can still execute if the callback fires

**Recommendation:** Audit all places where `transaction.on_commit()` or task queueing happens inside `tenant_schema_context()`, and ensure schema is passed as a parameter, not relying on the context being active.

---

### 4. **Throttling Rates May Be Too Generous for Public Booking Write**
**Why fix it:** Settings in `backend/backend/settings.py` define:
```python
'public_booking_write': config('PUBLIC_BOOKING_WRITE_THROTTLE_RATE', default='10/hour')
```
**Why fix it:** 
- 10 bookings per hour per IP allows potential abuse scenarios (scraping availability, DoS via booking spam)
- No mention of per-user throttling for authenticated members
- The default assumes a single prospect, but a team coordinator or bot could hit this limit legitimately

**Recommendation:** Make this configurable per environment and document the rate-limiting strategy in comments.

---

### 5. **ICS Feed "No Cancellation Permission" Logic is Weak**
**Why fix it:** `booking_tokens.py` issues two token types:
- `make_cancel_token()` — grants cancellation
- `make_feed_token()` — supposedly read-only

**Why fix it:** 
- The feed_token is signed with a different salt, BUT
- `read_feed_token()` falls back to accepting cancel tokens: `return _read_token(token, FEED_SALT) or read_cancel_token(token)`
- This means old cancel tokens work as subscription tokens, leaking write authority to the calendar feed endpoint

**Recommendation:** Remove the fallback and require fresh feed tokens on subscription requests.

---

### 6. **"Booking Group" Deduplication Assumes Consistent Metadata**
**Why fix it:** `prefer_visible_booking_copy()` deduplicates bookings using `metadata['booking_group']`. The logic:
```python
group = (event.metadata or {}).get("booking_group") or str(event.pk)
```
**Why fix it:**
- If metadata corruption occurs or a row is manually edited in the database, the group key is silently lost
- No validation that booking_group UUIDs are actually UUIDs
- A race condition during event creation could leave one sibling without the group, causing the card to appear twice

**Recommendation:** Add a database constraint or post_save signal to ensure booking_group is always set on paired events.

---

### 7. **Guest Privacy: Phone Numbers and Notes Leak to Organizer via Description Field**
**Why fix it:** When a guest books, `PublicBookingCreateView._event_description()` concatenates:
```python
lines = [f"Booked by {data['name']}", data["email"]]
if data.get("phone"):
    lines.append(data["phone"])
notes = (data.get("notes") or "").strip()
if notes:
    lines += ["", notes]
return "\n".join(lines)
```
**Why fix it:**
- This lands in `event.description`, which is visible to anyone who can see the event
- No access control on the description field prevents a stranger on a shared calendar from reading guest contact info
- Notes may contain sensitive information (e.g., "Calling to discuss salary negotiation")

**Recommendation:** Store phone, notes, and email on the guest EventAttendee row only, not in the description. Update the host's event dialog to display guest details from the attendee list, not the description.

---

### Summary
These issues span security (token fallback, privacy), reliability (CI fragility, dedup assumptions), and architectural concerns (PR scope, throttling). None appear to be show-stoppers for merging, but they are worth addressing either in follow-up PRs or in review comments before merge.

zizhenliu0427:
Thanks for the thorough review. I checked each point against the code at `c587660d0`. Two led to fixes on this branch; the rest are explained below.

## Fixed in this PR

**2. CI readiness check** — the stale-`.next` problem was tackled in three places. Yvonne's MED-314 (#800, still open) added a frontend restart after the in-container build; #808 independently added an equivalent restart and merged first, and #800 now carries that version after syncing with `prod-preview`. Those restarts wait for HTML only (`/health`, and in #808 also `/login`). This PR adds an asset-level check on top: after the last restart it loads the login page and fetches every `/_next/static` script it references, which catches a chunk mismatch an HTML-only check would miss. When merging `prod-preview` I kept #808's restart and wait steps and dropped this PR's own restart, so no extra restart is added, and the result merges cleanly with #800. The maintainability point about the inline Python is fair.

**7. Guest privacy** — agreed, and it went further than described. Team links book onto the project calendar, so every project member could read the guest's email, phone and notes, and the description is exported to the host's Google Calendar. Fixed in `fb060957b`:
- the event description no longer contains guest details (it now uses the booking link's own description);
- notes moved onto the guest's `EventAttendee` row, alongside email and phone;
- `EventAttendeeSerializer` blanks email, phone and notes for anyone other than the organiser and the guest;
- the event panel shows a guest section only when the API returns those details.

Moving the data onto the attendee row alone would not have been enough, because attendee rows use the same `view_all` permission as the event itself — so the redaction is enforced in the serializer.

**Related leak found while checking 5** — the confirmation page wrote the *cancel* link into the downloaded `.ics`, so a bearer token that can cancel the meeting ended up in guests' calendar apps. Fixed in `6be1550ae`: calendar entries now link to the booking page, which carries no token and lets a guest request a recovery link to their booking email. An E2E test asserts the downloaded `.ics` contains no cancel token.

## No change needed

**3. Rollback and task queueing** — `queue_booking_task` wraps `.delay()` in `transaction.on_commit`, and Django discards those callbacks on rollback. `tenant_schema` is passed explicitly to every task, and the recovery task resolves its own schema. The slot-conflict and Google-unavailable paths call `transaction.set_rollback(True)`.

**5. Feed token fallback** — the fallback lets a cancel token *read* the feed, not the other way round. The feed endpoint is read-only, and the cancel endpoint only accepts cancel-salted tokens. Removing the fallback would silently break subscriptions already saved in guests' calendars; legacy tokens expire after 400 days (`MAX_AGE_SECONDS`).

**6. `booking_group`** — both copies are written by `create_booking_events` inside a single `transaction.atomic` with one group UUID, so they commit together or not at all. `validate_metadata` rejects API changes to `booking_*` keys and `source`, and restores them when omitted. The group is only compared for equality, so its format doesn't matter, and a database constraint can't express a pairing across two rows.

## Proposed as follow-ups outside this PR

**1. Scope** — 6,494 of the 15,556 added lines are tests. After the merge the only CI change left is that asset-level check; the tenant-schema column sync in `core/services/tenant.py` is the other piece that isn't booking-specific.

**4. Throttling** — the rates are already environment-configurable (`PUBLIC_BOOKING_*_THROTTLE_RATE`), and signed-in bookers are keyed per user. The real problem is broader: `NUM_PROXIES` is unset, so DRF 3.14 identifies anonymous clients by the whole `X-Forwarded-For` header, and nginx's `$proxy_add_x_forwarded_for` appends to whatever the client sends. A spoofed header therefore gets a fresh rate-limit bucket on every request. That affects every throttle in the app and predates this PR, and the fix depends on the production proxy chain, so I'd raise it as its own ticket.

quanwangniuniu (changes requested):
Still need to fix those aftrer the optimization:

---

### **1. Multi-Tenant Schema Switching Without Explicit Rollback Protection**

**Problem:**  
Public booking endpoints manually call `tenant_schema_context()` to switch schemas. If an exception occurs during a booking operation:
- The `with` block exits and context restores, BUT
- Uncommitted transaction state is NOT guaranteed to be rolled back in worker queue tasks
- Tasks queued inside a failed atomic block can still execute if the callback fires

**Why Fix It:**  
This can cause orphaned tasks to run after the transaction fails, potentially creating calendar events or sending emails for bookings that were never actually confirmed. This leads to:
- Duplicate events on the calendar
- Emails sent for failed bookings
- Inconsistent state between database and external systems

**Fix:**  
Audit all places where `transaction.on_commit()` or task queueing happens inside `tenant_schema_context()` in `backend/calendars/booking_write.py` and `backend/calendars/booking_notifications.py`. Pass the schema as an explicit parameter to tasks rather than relying on context being active.

---

### **2. ICS Feed "No Cancellation Permission" Logic is Weak**

**Problem:**  
`booking_tokens.py` has a fallback in the feed token validation:
```python
return _read_token(token, FEED_SALT) or read_cancel_token(token)
```
This means old cancel tokens work as subscription tokens, leaking write authority to the calendar feed endpoint.

**Why Fix It:**  
- Security vulnerability: Cancel tokens should not grant read access to feeds
- A guest could reuse an old cancellation token to subscribe to future updates
- No clear separation between token permissions

**Fix:**  
Remove the fallback on line ~X in `booking_tokens.py`. Require fresh feed tokens exclusively for subscription requests. Never accept cancel tokens as valid feed tokens.

---

### **3. "Booking Group" Deduplication Assumes Consistent Metadata**

**Problem:**  
`prefer_visible_booking_copy()` deduplicates bookings using `metadata['booking_group']`:
```python
group = (event.metadata or {}).get("booking_group") or str(event.pk)
```

**Why Fix It:**  
- If metadata corruption occurs or a database row is manually edited, the group key is silently lost
- No validation that booking_group values are actually UUIDs
- A race condition during event creation could leave one sibling without the group, causing duplicate cards in the UI
- Silent fallback to `event.pk` masks the underlying corruption

**Fix:**  
- Add a database constraint to ensure `booking_group` is always set on paired booking events
- Add a post_save signal in `BookingEvent` model to validate and auto-populate missing booking_group values
- Add a migration check to backfill missing booking_group values on existing paired events

---

### **4. Guest Privacy: Phone Numbers and Notes Leak to Organizer via Description Field**

**Problem:**  
When a guest books, phone number, email, and notes are concatenated into `event.description`:
```python
lines = [f"Booked by {data['name']}", data["email"]]
if data.get("phone"):
    lines.append(data["phone"])
```

**Why Fix It:**  
- This lands in `event.description`, visible to anyone who can see the event
- No access control prevents strangers on a shared calendar from reading guest contact info
- Notes may contain sensitive information (e.g., salary negotiation details)
- Violates guest privacy expectations for anonymous bookings

**Fix:**  
- Remove phone, email, and notes from the event description in `PublicBookingCreateView._event_description()`
- Store phone, notes, and email on the `EventAttendee` row only
- Update the host's event dialog in `EventPanelDialog.tsx` to display guest details from the attendee list, not the description
- Ensure attendee details are only visible to the event owner, not to other calendar viewers

---

zizhenliu0427:
Thanks for the second pass. Three things changed, and two I've left as they are with reasons below. Everything here sits on top of `2baa96c35`, the commit the review was made against.

## Changed

**Feed tokens.** `read_feed_token` no longer falls back to accepting a cancel token — only a feed token opens the feed. A cancel token now gets the same 404 as a forged one, and a feed token still cannot cancel. Unit tests cover both directions, and the feed tests that previously passed a cancel token now use the real subscription token from `feed_url`. Nothing has shipped yet, so the only subscription URLs this invalidates are from QA runs.

**Booking group.** A booking row without a group now logs a warning instead of quietly degrading: in `booking_siblings`, where a missing group means the other copy will not be cancelled or moved with it, and in the recovery dedupe. Ordinary events do not log, since they never carry a group. Tests cover both.

I have not added the database constraint or the `post_save` backfill. The pairing spans two rows and a `CHECK` constraint is per-row, so it cannot express "these two share an id". A signal cannot repair a lost pairing either — it could only mint a new id, which would create a *wrong* grouping rather than a missing one. The warning is the part that turns silent degradation into something visible.

**Rollback.** Added a regression test: when a booking is rejected because the slot was taken, no task is queued.

## Not changed

**Guest privacy.** This was fixed in `fb060957b`, which is an ancestor of `2baa96c35`. The code quoted in the review no longer exists: `_event_description(link)` returns only the booking link's own description. Guest email, phone and notes are stored on the `EventAttendee` row, and `EventAttendeeSerializer` blanks them for anyone who is neither the organiser nor the guest. The event panel shows a guest section only when the API returns those details.

One detail from the suggested fix is worth flagging: storing the details on the attendee row alone does not restrict who can read them, because attendee rows are gated by the same `view_all` event permission as the event itself. That is why the redaction is enforced in the serializer.

**Task queueing.** `queue_booking_task` wraps `.delay()` in `transaction.on_commit`, and Django discards those callbacks when the transaction rolls back, so a failed booking leaves no task behind. `tenant_schema` is already passed to every task explicitly rather than relying on the active context, which is what the review asks for, and the recovery task opens its own schema. `booking_write.py` and `booking_notifications.py` do not queue tasks at all. The slot-conflict and Google-unavailable paths also call `transaction.set_rollback(True)`.

Backend tests for the affected areas: 100 passed. CI was green on `2baa96c35`; I will re-run it with these commits.

quanwangniuniu (approved):
Good job

PR #807, opened 2026-09-09
fix(auth): improve registration password feedback

PianoMaksim:
Improve registration password feedback so users can see why a password is invalid before submitting.

- Show a concise password requirement summary when the password field is empty.
- Replace the summary with live validation feedback while typing: ❌ with specific error reasons, or ✅ “Password is valid.”
- Refresh validation when the password, username, or email changes.
- Display backend password validation errors beneath the password field if registration fails.
- Reuse the existing Django password validators without adding new password rules. Keep feedback in English.
- Add a password validation endpoint and update the OpenAPI documentation.

### DIFF
--- a/backend/authentication/serializers.py
+++ b/backend/authentication/serializers.py
@@ -8,6 +8,22 @@ from stripe_meta.models import Subscription
 
 User = get_user_model()
 
+
+class PasswordValidationStringField(serializers.Field):
+    default_error_messages = {'invalid': 'Not a valid string.'}
+
+    def to_internal_value(self, data):
+        if not isinstance(data, str):
+            self.fail('invalid')
+        return data
+
+
+class PasswordValidationSerializer(serializers.Serializer):
+    password = PasswordValidationStringField(default='')
+    username = PasswordValidationStringField(default='')
+    email = PasswordValidationStringField(default='')
+
+
 class OrganizationSerializer(serializers.ModelSerializer):
     plan_id = serializers.SerializerMethodField()
     
--- a/backend/authentication/urls.py
+++ b/backend/authentication/urls.py
@@ -3,6 +3,7 @@ from django.urls import path
 from rest_framework_simplejwt.views import TokenRefreshView
 from .views import (
     RegisterView,
+    PasswordValidationView,
     VerifyEmailView,
     LoginView,
     LogoutView,
@@ -23,6 +24,7 @@ from .views import (
 
 urlpatterns = [
     path('register/', RegisterView.as_view(), name='register'),
+    path('password/validate/', PasswordValidationView.as_view(), name='password-validate'),
     path('verify/', VerifyEmailView.as_view(), name='verify'),
     path('login/', LoginView.as_view(), name='login'),
     path('logout/', LogoutView.as_view(), name='logout'),
--- a/backend/authentication/views.py
+++ b/backend/authentication/views.py
@@ -8,10 +8,14 @@ from rest_framework.permissions import IsAuthenticated
 from rest_framework.parsers import MultiPartParser, FormParser, JSONParser
 from django.contrib.auth import get_user_model
 from django.contrib.auth import authenticate
-from django.contrib.auth.password_validation import validate_password
+from django.contrib.auth.password_validation import (
+    UserAttributeSimilarityValidator,
+    get_default_password_validators,
+    validate_password,
+)
 from django.core.exceptions import ValidationError
 from rest_framework_simplejwt.tokens import RefreshToken
-from .serializers import UserProfileSerializer, OrganizationTokenRefreshSerializer
+from .serializers import UserProfileSerializer, OrganizationTokenRefreshSerializer, PasswordValidationSerializer
 from .login_security import LoginSecurityService
 from .password_rotation import get_password_rotation_status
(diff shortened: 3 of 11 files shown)


quanwangniuniu (changes requested):
---

## **Issue 1: Missing Input Validation in `PasswordValidationView`**

**File:** `backend/authentication/views.py` (lines 78–110)

**Reason:**
The `PasswordValidationView` uses a custom `PasswordValidationStringField` that only validates the field type, but it doesn't sanitize or validate the actual content of the strings. This creates security and stability risks:

- **XSS Risk (Frontend):** The `help_text` and `errors` from Django validators are returned verbatim in JSON and rendered in the browser. If a custom Django password validator injects unsanitized HTML or JavaScript, it could execute in the frontend.
- **No Length Limits:** The endpoint accepts unbounded strings for `password`, `username`, and `email`. An attacker could send extremely large payloads to cause performance issues or OOM errors.
- **Incomplete Validation:** The custom field allows empty strings by default (`default=''`), which means all three fields can be omitted, but there's no validation that the data makes sense (e.g., username should not contain only whitespace).

**How to Fix:**
```python
class PasswordValidationStringField(serializers.Field):
    def __init__(self, *args, max_length=1000, **kwargs):
        super().__init__(*args, **kwargs)
        self.max_length = max_length
    
    def to_internal_value(self, data):
        if not isinstance(data, str):
            self.fail('invalid')
        if len(data) > self.max_length:
            self.fail('max_length')
        return data.strip()  # Sanitize whitespace

class PasswordValidationSerializer(serializers.Serializer):
    password = PasswordValidationStringField(default='', max_length=256)
    username = PasswordValidationStringField(default='', max_length=150)
    email = PasswordValidationStringField(default='', max_length=254)
```

Also, escape or sanitize validator messages before returning them:
```python
import html
rules.append({
    'id': validator.__class__.__name__,
    'help_text': html.escape(help_text),  # Escape HTML entities
    'valid': not errors if password else None,
    'errors': [html.escape(e) for e in errors],
})
```

---

## **Issue 2: Race Condition in Password Validation on Frontend**

**File:** `frontend/src/components/auth/PasswordRequirements.tsx` (lines 23–47)

**Reason:**
The component fetches validation results asynchronously with a 300ms debounce, but there's a **race condition** that can cause stale validation results to display:

1. User types password → `inputKey` changes → fetch starts
2. User quickly changes password again → new `inputKey` → new fetch starts
3. If the second fetch completes *before* the first, `inputVersion` is updated to reflect the new input, but `validation.inputKey` still points to old input state
4. The component then renders validation for the old input, which is now incorrect

Additionally, the component doesn't handle **network errors gracefully**—if the validation endpoint is down, users see "Password check unavailable" but can still submit, which defeats the purpose of live validation.

**How to Fix:**
```typescript
export default function PasswordRequirements({ password, username, email }: Props) {
  const inputKey = JSON.stringify([password, username, email]);
  const [validation, setValidation] = useState<{
    inputKey: string;
    rules: PasswordValidationRule[];
    unavailable?: boolean;
  } | null>(null);

  useEffect(() => {
    const controller = new AbortController();
    const timer = setTimeout(async () => {
      try {
        const result = await authAPI.validatePassword(
          { password, username, email },
          controller.signal,
        );
        // Only update if this is the latest request
        setValidation(prev => 
          prev?.inputKey === inputKey ? { inputKey, rules: result.rules } : prev
        );
      } catch (error) {
        if (controller.signal.aborted) return; // Request was cancelled
        setValidation(prev => ({
          inputKey,
          rules: prev?.rules || [],
          unavailable: true,
          error: error instanceof Error ? error.message : 'Unknown error',
        }));
      }
    }, password ? 300 : 0);

    return () => {
      clearTimeout(timer);
      controller.abort();
    };
  }, [inputKey]); // Depend on inputKey, not individual fields

  // Only render rules if they match the current input
  const current = validation?.inputKey === inputKey;
  // ... rest of component
}
```

---

PianoMaksim:
Thanks for the review. Here are my responses to the five points:

XSS: I did not add backend html.escape() because the frontend already renders these messages as plain text. Additional escaping would cause entities such as &lt; to appear literally, creating display issues. {rule.errors.join(' ')} in PasswordRequirements and {error} in FormInput are both JSX text children. React renders these strings as text without interpreting HTML tags or executing event-handler code, so I retained the existing implementation.

Length limits: Enforced maximum lengths of 256 characters for passwords, 150 for usernames, and 254 for emails, with corresponding error messages. The password precheck and registration endpoint share the same validation.

Whitespace handling: Usernames and emails are trimmed, and whitespace-only values are rejected. Passwords remain unchanged to preserve the user's input, since some users may intentionally include leading or trailing spaces.

Race conditions: When the input changes, the cleanup function in PasswordRequirements sets active = false. Both success and error callbacks check if (active) to prevent outdated results from updating the state. Before displaying feedback, validation.inputKey === inputKey also ensures the result matches the current input. Out-of-order response tests passed, so I retained this logic.

Network errors: Registration is disabled while the check is pending or unavailable. Failures display an unavailable message and a Retry button. Once a retry returns a validation result, the existing submission logic is restored.

The changes and relevant tests have been completed locally.

quanwangniuniu (approved):
Good job

PR #809, opened 2026-09-09
MED-264: Add RAG retrieval over project documents

JadeShi-1018:
## Summary

Adds project-scoped RAG retrieval over project documents for MED-264.

The Agent can now retrieve grounded context from:
- Meetings / MeetingDocument
- Notion Draft / ContentBlock
- Retrospectives

## Implementation

- Added new `rag` app with document extraction, chunking, embeddings, incremental indexing and exact cosine retrieval
- Added `DocumentChunk` and `DocumentIndexState`
- Added explicit `DraftProjectLink` for project-scoped Draft indexing
- Added indexing signals for meetings, Notion content and retrospectives
- Integrated project RAG into the Agent fallback path
- Added grounded citations and persisted citation metadata
- Added citation rendering/restoration in the Agent UI
- Added tenant-safe handling for the streaming response lifecycle
- Corrected tenant provisioning for retrospective project-scoped models
- Added pgvector setup for embedding storage

## Evaluation

Canonical retrieval eval dataset:
- 58 sources
- 44 questions

Local BGE baseline:

**Single-source**
- Hit@1: 0.917
- Hit@3: 0.958
- Hit@5: 0.958
- MRR: 0.944

**Multi-source**
- Recall@3: 0.817
- Recall@5: 0.900
- Recall@10: 0.933
- Full-Coverage@3: 0.600
- Full-Coverage@5: 0.733
- Full-Coverage@10: 0.800

The local BGE provider is included as a local/eval fallback while the current dev Gemini embedding key is unavailable. Gemini remains the default provider.

## QA

- Project RAG integration tests: PASS
- Streaming tenant-context regression tests: PASS
- MED-264 Playwright `ask about my retrospective` flow: PASS
  - grounded answer rendered
  - Sources rendered
  - Retrospective citation rendered
  - citations persisted after reload / reopening the message board

## Seed Data

Added:

`backend/rag/eval_seed/med264_seed.sql`

The SQL seeds canonical source/domain data only. Chunks, index state and embeddings are still generated through the real indexing path.


quanwangniuniu (changes requested):
---

## **Problem 1: Incomplete Tenant Safety Implementation**
**Why fix it:** The PR adds tenant-scoped RAG models to `core.tenant_config.get_tenant_models()` but leaves `RetrospectiveTask` and `Insight` in the public schema instead of the tenant schema. This creates cross-schema join complications (RetrospectiveTask in public, DocumentChunk in tenant) that aren't fully resolved yet — see the comment in `core/tenant_config.py` marking this as "MED-264 tenancy fix" but still incomplete. A future change will be required to move retrospectives fully into tenant scope, creating technical debt.

---

## **Problem 2: Streaming Response Tenant Context Bug (Critical)**
**Why fix it:** The `ChatView` wraps `event_stream()` in a generator before `TenantSchemaMiddleware`'s `finally` block resets `search_path` to `public`. Every RAG retrieval call inside the generator runs against the empty public schema unless the schema is re-captured and re-entered inside the generator itself. The test in `test_chat_view_tenant_streaming.py` shows this was silently broken until fixed, with every question falling into the "no chunks found" fallback. **This is a data isolation/multi-tenant correctness issue.**

---

## **Problem 3: Embedding Provider Coupling Without Provider Abstraction Layer**
**Why fix it:** Both `DocumentChunk.embedding` and the Gemini embedding call assume a fixed 768-dimension vector because `gemini-embedding-2` is hardcoded as the default. The PR includes a temporary `RAG_EMBEDDING_PROVIDER=local` fallback for eval, but this is documented as "never production." **Switching providers later (e.g., to OpenAI or Anthropic) requires re-embedding every DocumentChunk row, which is not automated and will silently corrupt retrieval quality.** The PR should enforce provider consistency through a migration guard or a explicit versioning strategy.

---

## **Problem 4: Missing Similarity Threshold Tuning & No Default Min Similarity**
**Why fix it:** `RAG_RETRIEVAL_MIN_SIMILARITY` defaults to `None` (no threshold), so `retrieve_chunks()` always returns top-K even for low-confidence matches. The `answer_project_question()` LLM prompt says "If the retrieved context does not support the answer, say so," but it has no way to know when context is too weak — the decision is deferred to the LLM, which introduces hallucination risk. The eval dataset doesn't provide empirical data on a good threshold, leaving retrieval quality tuning to future work, **which risks deploying with noisy/irrelevant results to production users.**

---

## **Problem 5: Incomplete Citation Metadata & No Real-Time Rendering Verification**
**Why fix it:** The PR persists `citation_metadata` (title, meeting_id, scheduled_date, etc.) but the frontend UI code for rendering citations is not shown in this PR. `answer_project_question()` yields citations in a structured format, but without verifying the frontend actually reconstructs them correctly from that structure, citations could silently fail to render or render incorrectly — e.g., links break, dates aren't formatted properly, or metadata fields are missing. **The Agent UI integration (the "citation rendering/restoration" listed in the PR summary) needs explicit test coverage.**

---

## **Problem 6: No Rollback Plan for Index State Corruption**
**Why fix it:** If `DocumentIndexState` rows become inconsistent (e.g., `status=COMPLETE` but zero `DocumentChunk` rows exist), the `_resolve_unchanged_state()` function logs a warning and forces a full re-index. However, there's no automated recovery mechanism, no alert/monitoring, and no ops runbook for what to do if many projects' index states are corrupted. **In production, a silent database inconsistency could leave users with stale/incomplete retrieval results indefinitely if nobody notices the warning log.**

---

## **Problem 7: Signal-Based Enqueuing Has Best-Effort Delivery Only**
**Why fix it:** The RAG indexing signals (in `meetings/signals.py`, `notion_editor/signals.py`, etc.) use `transaction.on_commit()` + `rag.tasks.index_document_task.delay()`, which has two separate failure points: (1) DB commit succeeds but Celery enqueue fails (broker outage), or (2) the enqueue callback itself raises but is logged and swallowed. **There is no transactional outbox or guarantee that every source change gets indexed.** The PR says "recovery is via rebuild_project_index," but an ops person has to notice a source went unindexed and manually trigger the rebuild — no automatic replay or detection.

---

## **Problem 8: No Explicit Eval Results in the PR Description**
**Why fix it:** The PR lists eval metrics (Hit@1: 0.917, Recall@3: 0.817, etc.) but doesn't include:
- The actual **eval_results JSON artifact** from running `run_project_rag_eval`
- A breakdown of which **question types** pass/fail (single-source success rate vs. multi-source)
- Any analysis of **failure modes** (e.g., which distractors confuse the retriever most)
- Reproducibility: can someone re-run the eval locally and get the same numbers?

**Without this**, reviewers can't independently verify the baseline quality claims, and there's no clear acceptance criteria for "good enough" retrieval — the threshold for shipping to production is undefined.

---

JadeShi-1018:
Thanks for the detailed review. I went through each point individually and verified the current implementation before making changes.

A few comments exposed real gaps and resulted in code changes, especially around embedding-pipeline compatibility and durable recovery for missed/failed indexing work. Some concerns were already covered by the existing implementation, and for others I intentionally kept the current behavior where the available eval data did not justify a stronger mechanism.

The main updates in this revision are:

- Strengthened embedding provider/pipeline invalidation so incompatible vector spaces cannot be mixed during retrieval.
- Added durable RAG indexing recovery with `dirty_since`, synchronous cleanup for confirmed exclusions/moves, and a periodic reconciler for missed Celery publishes or unsuccessful indexing attempts.
- Added regression coverage for stale/duplicate workers, source moves, delete/exclusion races, project isolation, and reconciler behavior.
- Added a tracked, reproducible local-BGE evaluation baseline plus documentation for reproducing the eval and inspecting per-question failures.

I also reviewed the points where I did not make a production change (for example, the similarity threshold and some operational recovery tooling) and documented the trade-offs and possible follow-up work rather than expanding this PR unnecessarily.

I’ll reply to each review point separately with the exact change, effect, remaining limitation, and possible follow-up.

---

## 1. Tenant safety / Retrospective & Insight scoping

**Status:** Addressed with a small correctness/test update.

I verified the current tenant registration rather than assuming the older comments were still accurate. `RetrospectiveTask`, `CampaignMetric`, and `Insight` are already tenant-scoped models in the current configuration, so retrospective RAG data is not intentionally stored in `public`.

### Changes

- Updated stale comments that still described retrospective/insight data as public-scoped.
- Added a regression test asserting that these models remain registered in the tenant schema configuration.

### Effect

New retrospective/insight writes follow the same tenant isolation model as the other RAG sources, and the regression test protects this assumption from silently changing.

### Not done

I did **not** add a historical production-data backfill from `public` into tenant schemas.

The repository cannot establish whether an older deployment actually contains legacy rows there, and blindly moving rows without knowing their tenant ownership would be unsafe.

### Possible follow-up

If legacy production data is discovered, I would handle that as a deployment migration/runbook task:

1. inspect existing rows;
2. determine their tenant ownership;
3. migrate them explicitly into the correct schemas.

### Trade-off

This PR guarantees correctness of the current schema/configuration without attempting a speculative historical data migration.

---

## 2. Tenant context during streaming responses

**Status:** Already addressed in the current implementation; no additional production change was required during this review pass.

The streaming view captures the active tenant schema while the request/middleware context is still active, and the SSE generator explicitly re-enters that schema for the lifetime of the stream.

This matters because the generator executes after the original request stack may have exited. Relying on middleware connection state alone could therefore lose tenant context during streaming.

### Verification

Tests cover:

- normal streaming;
- error handling;
- early generator close.

### Effect

All DB work performed while generating the streamed agent response remains inside the originating tenant schema.

### Not done

I did **not** introduce a generic streaming-context framework or refactor unrelated streaming endpoints.

### Possible follow-up

A shared helper could be extracted later if the same pattern is required by additional streaming views.

### Trade-off

The fix stays local to this RAG/agent flow rather than broadening MED-264 into a framework-level refactor.

---

## 3. Embedding provider / pipeline compatibility

**Status:** Changed.

I made the embedding provider/model part of the active indexing pipeline identity rather than relying only on the configured model name.

The pipeline hash now reflects:

- embedding provider/model identity;
- embedding dimensions;
- chunking configuration;
- document/query framing;
- index version.

Retrieval also checks the source's `DocumentIndexState` against the **current pipeline hash**, so chunks generated by an older or incompatible embedding pipeline are excluded.

### Effect

Switching, for example, from Gemini embeddings to local BGE cannot silently mix vectors from two different embedding spaces.

Existing stale vectors become ineligible for retrieval, and the source is re-indexed when normal indexing/rebuild is triggered.

### Intentional behavior

I intentionally did **not** gate retrieval on:

```python
status == COMPLETE
````

A `FAILED` or `PENDING` indexing attempt can coexist with valid last-known-good chunks from the **current** pipeline, so status alone is not a safe proxy for vector validity.

### Not done

I did **not** add a fleet-wide automatic rebuild immediately triggered by an embedding configuration change.

Existing rebuild/indexing paths remain responsible for materializing the new vectors.

### Verification

Focused tests cover:

* provider-switch invalidation;
* stale/current mixed states;
* `NULL` state compatibility;
* missing state compatibility;
* provider-switch indexing behavior.

### Trade-off

Stale vectors are prevented from contaminating retrieval immediately, while re-embedding remains an explicit/lazy operational action rather than adding global configuration-change orchestration.

---

## 4. Similarity threshold

**Status:** Reviewed and intentionally not changed.

I evaluated the current local-BGE dataset before introducing a default cosine-similarity threshold.

The score distributions overlap too much for a single global threshold to be safe:

* no-answer questions reached roughly `0.67` Top-1 similarity;
* legitimate evidence for difficult answerable/multi-source questions appears around `0.58–0.66`.

A threshold high enough to suppress some no-answer retrievals would therefore also remove evidence required for valid answers.

### Current behavior

The default remains:

```python
RAG_RETRIEVAL_MIN_SIMILARITY = None
```

The system instead relies on:

* Top-K retrieval;
* grounded prompting;
* answer abstention when the retrieved evidence is insufficient.

### Effect

Retrieval recall is preserved rather than trading valid evidence away for an uncalibrated rejection rule.

### Not done

I did **not** derive a production Gemini threshold from local BGE scores because absolute cosine values are embedding-provider dependent.

### Possible follow-up

A future calibration pass could use:

* a larger hard no-answer set;
* full per-rank diagnostics;
* a production-Gemini evaluation baseline.

### Trade-off

The current design favors recall and grounded abstention over aggressive similarity filtering until there is enough provider-specific data to calibrate a threshold safely.

---

## 5. Citation persistence / frontend visibility

**Status:** Already implemented; no production change was required for this feedback.

Retrieval returns structured citation metadata including:

* source type / source ID;
* chunk index;
* title metadata;
* snippet;
* similarity;
* citation number.

That metadata flows through:

```text
retrieval
→ SSE response
→ assistant metadata
→ AgentMessage.metadata
→ restored chat history
```

The frontend renders the citations as source/evidence cards.

### Verification

The Playwright flow verifies that the relevant retrospective source appears:

* in the live response;
* after reload/reopen.

### Effect

Citations are not ephemeral UI state. They survive persistence and can be inspected after reopening the conversation.

### Not done

I did **not** add clickable deep-links into the underlying Meeting/Draft/Retrospective pages because that was not part of the MED-264 acceptance criteria.

### Possible follow-up

Potential follow-ups include:

* deep-link navigation to source records;
* more isolated frontend citation unit tests.

### Trade-off

This PR delivers evidence provenance and persistence without expanding the ticket into source-navigation UX.

---

## 6. Index recovery / corruption / rollback

**Status:** Existing recovery behavior was audited; no large recovery subsystem was added for this item.

The RAG index is treated as derived data from the business sources.

Successful chunk/state reconciliation is:

* atomic;
* protected by row locking.

If a later re-index fails, the existing last-known-good chunks/hashes are preserved rather than being partially replaced.

The indexer also detects a state that claims to be current but has zero chunks and forces re-indexing.

Retrieval additionally excludes sources whose pipeline hash is stale.

A project-level rebuild path already exists and re-enumerates the live business sources.

### Effect

Normal failure modes do not leave a partially-written current index, and common derived-state inconsistencies can either:

* self-heal during indexing; or
* be repaired by rebuilding the project index.

### Not done

I did **not** add:

* an admin repair UI;
* a fleet-wide repair command/runbook;
* an orphan sweeper for rows created by bypassing normal ORM/signals with raw SQL.

### Possible follow-up

Those would be useful operational hardening items if explicit production repair tooling becomes necessary.

### Trade-off

Because the vector index is fully derived from business-source data, this PR keeps recovery source-of-truth based rather than building a second transactional/versioned storage system around the derived chunks.

---

## 7. Signal reliability / lost Celery publishes

**Status:** Changed substantially.

I added a dedicated durable `dirty_since` marker to `DocumentIndexState`.

Source signals synchronously mark the logical RAG source dirty before scheduling asynchronous indexing.

`status` remains the lifecycle of an indexing attempt and is deliberately separate from whether source reconciliation is still owed.

### Reconciliation behavior

A successful current reconciliation clears:

```text
dirty_since
```

Failed or superseded attempts do **not** clear it.

I also added a periodic:

```python
reconcile_dirty_rag_states
```

watchdog.

It scans dirty rows older than the configured threshold across tenant schemas and re-enqueues the existing `index_document_task`, regardless of whether the last status is:

* `COMPLETE`;
* `PENDING`;
* `FAILED`.

An individual broker/publish failure is logged and does not stop other dirty rows from being attempted. The marker remains in place for the next periodic run.

### Delete / exclusion / move handling

For delete, exclusion, and move-away cases, I also added a synchronous DB-only cleanup fast path.

Before clearing anything it re-extracts the current source while holding the RAG state lock.

This protects against stale cleanup calls deleting a source that has since become valid again, including the Draft:

```text
A → B → A
```

ownership race.

### Effect

A lost Celery publish or unsuccessful indexing attempt no longer depends on somebody editing the source again later for recovery.

Marked mutations are automatically retried by the periodic reconciler.

Delete/exclusion operations also avoid leaving stale chunks visible while waiting for Celery.

### Verification

Focused verification is green:

* **7** reconciler tests;
* **47** indexing/signal tests;
* migration/model consistency checks.

### Not done

I did **not** implement:

* a transactional outbox;
* broad outer transactions around unrelated source-writing code.

There remains a very narrow autocommit failure window where the source SQL write could commit and the process could die before the synchronous dirty marker is written.

Eliminating that window completely would require stronger transaction/outbox integration with the business write path.

### Possible follow-up

If stronger delivery guarantees become necessary, the next step would be a transactional outbox tied directly to the source write transaction rather than adding additional retry layers to Celery.

### Trade-off

`dirty_since` + periodic reconciliation provides durable, bounded recovery for marked mutations with much less infrastructure than a transactional outbox.

The system remains **eventually consistent**, rather than strongly consistent.

Duplicate re-enqueues are intentionally tolerated because the existing indexing path is idempotent and concurrency-safe.

---

## 8. Eval reproducibility / baseline artifact

**Status:** Changed at the documentation/artifact level; the eval implementation itself did not require redesign.

The existing eval runner already produces:

* aggregate retrieval metrics;
* per-question diagnostics;
* expected sources;
* retrieved Top-5 sources;
* rank;
* similarity;
* snippets;
* outcome;
* categorized failure reasons.

The real gap was discoverability and reproducibility.

Generated results previously lived under the gitignored:

```text
backend/rag/eval_results/
```

so reviewers could not inspect the actual evaluation output from the PR.

### Changes

Added:

* `backend/rag/eval_seed/README.md`

  * reproducible local-BGE command;
  * fixture/SQL workflow;
  * metric definitions;
  * dataset composition;
  * known partial cases.

* `backend/rag/eval_seed/baselines/med264_v2.1.0_local_bge_baseline.json`

  * a tracked, byte-identical copy of the real canonical evaluation result.

The canonical dataset contains:

* **58 sources**
* **44 questions**

  * 24 single-source;
  * 15 multi-source;
  * 5 no-answer.

### Effect

A reviewer can now inspect the actual MED-264 retrieval result without running the harness.

Future retrieval changes also have a fixed reference point for comparing:

* aggregate metric deltas;
* per-question regressions/improvements.

### Known baseline gaps

The baseline is intentionally **not** described as fully passing.

It preserves the current three partial multi-source cases:

* `mq06`;
* `mq13`;
* `mq14`.

Their failure classifications remain visible rather than being hidden from the baseline.

### Not done

I did **not**:

* commit every timestamped eval output;
* add new metrics;
* change retrieval behavior;
* modify the eval runner.

Temporary/generated runs remain gitignored.

### Reproducibility choice

The committed baseline uses deterministic local BGE rather than Gemini so the reference can be reproduced without API spend or external-provider variability.

### Possible follow-up

Future improvements could include:

* automated CI/regression comparison against the tracked baseline;
* a separately recorded production-Gemini baseline.

### Trade-off

The committed local-BGE baseline optimizes for reproducibility and stable comparison.

It is not intended to claim that its absolute cosine values are identical to production Gemini behavior.

PR #811, opened 2026-09-12
[core] Extract Decision backend into .NET microservice

qiwangai:
## Summary
- Add a standalone .NET Decision service with controllers, JWT validation, project permission checks, Postgres repository, and API contracts.
- Route /api/decisions through nginx and the Next.js API proxy to the new decision-service container.
- Remove the old Django Decision API views/urls/tests while keeping shared models and migrations for existing table ownership during migration.
- Move Decision graph/topic label calls to the .NET service and fix flat/nested project routes to send numeric project ids.


qiwangai:
I extracted the Decision backend out of the Django monolith into a standalone .NET service.

The frontend API path stays the same (`/api/decisions`), but requests are now routed through nginx / frontend proxy to the new `decision-service`. The .NET service owns the Decision API logic, connects directly to Postgres, validates Django-issued JWTs, and performs project membership / role-based permission checks before returning or mutating Decision data.

Main changes:
- Added `services/decision-service` as a new .NET backend service.
- Moved Decision API handling into .NET controllers.
- Added Postgres repository logic for existing Decision tables.
- Added JWT validation and project permission checks inside the .NET service.
- Added docker compose support for the new service.
- Routed `/api/decisions` traffic from nginx and frontend proxy to the .NET service.
- Removed the old Django Decision API routes/views/tests to avoid duplicate ownership.
- Added .NET tests for API behavior, JWT validation, and permission policy.

The frontend behavior should remain unchanged. This is mainly a backend ownership/architecture change: Decision API is now served by .NET instead of Django.

quanwangniuniu (approved):
Good job

PR #812, opened 2026-09-13
fix(api): classify HTTP client timeouts by endpoint duration (MED-399)

wenxiaogao000:
## Summary
Replaces the single global 10s axios timeout plus ad-hoc per-module overrides
with named tiers exported from `lib/api.ts`: `LLM_TIMEOUT_MS`,
`LLM_BATCH_TIMEOUT_MS`, `SYNC_TIMEOUT_MS`, `UPLOAD_TIMEOUT_MS`, and
`LONG_REQUEST_TIMEOUT_MS`.

## What changed
- Existing overrides (pattern/pivot/quickStart/spreadsheet/googleCalendar)
  now reference the shared constants instead of duplicating local numbers.
- Adds the two previously-missing overrides: chat attachment upload
  (`UPLOAD_TIMEOUT_MS`) and the Facebook Graph API sync call (`SYNC_TIMEOUT_MS`).
- Fixed a test regression found while verifying this: `jest.mock('@/lib/api', ...)`
  was replacing the whole module, silently dropping the new named exports
  under test.

## Investigated but did not change
`agentApi.ts`, `reportApi.ts`, `optimizationApi.ts`, and
`optimizationScalingApi.ts` — the ticket also names these, but their
axios-mediated calls are plain CRUD. The genuinely slow agent interactions
(chat, upload-analyze) already bypass axios via streaming fetch, report-task
creation does no generation inline, and campaign optimisation syncs run as
scheduled Celery tasks decoupled from any HTTP request.

## Out of scope
The job-and-poll direction the ticket also suggests for genuinely long
operations depends on the Celery worker time-limit fix (MED-400) landing
first, per Ray.

## Testing
- `npx tsc --noEmit` — passes
- `npx jest src/__tests__/lib/api` — 34 passed
- `npm run build` — passes, all routes compile

🤖 Generated with [Claude Code](https://claude.com/claude-code)


quanwangniuniu (approved):
Good job! But next time remember QA first

PR #814, opened 2026-09-15
fix(auth): end session when refresh token is rejected + MED-397

zizhenliu0427:
## Summary

Fixes MED-397 — expired session never logs the user out and retries forever. When the server rejects a refresh token (for example, because it has expired), the frontend now clears the stored session and redirects to `/login`. Previously, the client silently swallowed the failure, left the user stranded on a broken-looking page, and issued a fresh refresh call for every subsequent API request.

- **Session death on rejection** — a new `endSession()` in `api.ts` clears persisted auth state, fires an `auth:session-ended` event so the Zustand store resets in-memory state, and hard-redirects to `/login`. It runs once per shared refresh, not once per waiting request.
- **Retryable vs non-retryable** — `getSharedRefreshedToken` now uses the raw `requestTokenRefresh` (which throws on failure) instead of `refreshAccessToken` (which catches everything and returns null). A 4xx ends the session; a 5xx or network error keeps the user signed in so a backend blip does not force everyone to re-authenticate.
- **permissionApi single-flight** — `permissionApi.ts` was calling `refreshAccessToken` directly, bypassing the shared de-duplication. It now uses `getSharedRefreshedToken`, so concurrent 401s across all three HTTP clients share one refresh call.
- **Google Docs exception preserved** — the `shouldBypassGlobalLogout` guard in `api.ts` is unchanged. A Google Docs integration 401 still passes through without triggering refresh or logout.

## Files Changed

(5 files)

**Frontend — API client layer**
- `frontend/src/lib/api.ts` — extract `requestTokenRefresh` (throws); add `endSession()`, `SESSION_ENDED_EVENT`; `getSharedRefreshedToken` catches and calls `endSession` on non-retryable failure; interceptor calls `endSession` when no refresh token exists but a bearer was sent
- `frontend/src/lib/api/preferencesApi.ts` — import `endSession`; interceptor calls `endSession` when no refresh token exists but a bearer was sent
- `frontend/src/lib/api/permissionApi.ts` — switch from `refreshAccessToken` to `getSharedRefreshedToken`; import `endSession`; call `endSession` when no refresh token exists but a bearer was sent
- `frontend/src/lib/authStore.ts` — listen for `SESSION_ENDED_EVENT` and call `clearAuth()` to reset in-memory auth and chat state

**Frontend — tests**
- `frontend/src/__tests__/lib/tokenRefreshSingleFlight.test.ts` — expand "refresh fails" test to assert session is cleared; add "server error keeps session" test; add "Google Docs 401 bypasses refresh and logout" test

## Expected Result

- A rejected refresh token (401/400 from `/auth/token/refresh/`) clears the session and redirects to `/login`.
- A server error (5xx) or network failure on refresh keeps the user signed in.
- Concurrent 401s across `api`, `preferencesApi`, and `permissionApi` share one refresh call (existing behaviour, now also covering permissionApi).
- Google Docs integration 401s do not trigger refresh or logout (existing behaviour, preserved).
- Opening a public share link while the browser holds a dead session sends the user to `/login`.

## Follow-up Recommendation

- **Backend token revocation:** the refresh endpoint does not check `auth_token_version`, so after a password change it keeps issuing access tokens that are immediately rejected — other tabs and devices get stuck with failing requests instead of being logged out. Logout also does not revoke refresh tokens, because the SimpleJWT token blacklist app is not installed. Both need backend fixes; once the refresh endpoint rejects those tokens, this change logs those sessions out.
- **Two-tab logout:** other open tabs are not notified when one tab logs out. A `BroadcastChannel` or `storage` event listener could propagate the logout immediately.
- **Portal client:** `portalAuth.ts` has its own refresh flow for the client portal. It has the same gap but is a separate login system and out of scope for this ticket.

## Test Plan

- [x] `tokenRefreshSingleFlight.test.ts` — **5 passed** (3 existing tests, one extended to assert the session is cleared, + 2 new)
- [x] Frontend lint — clean on all 5 changed files
- [x] `tsc` — only the 5 pre-existing e2e spec errors
- [x] Manual QA: sign in → replace tokens with junk → click around → expect one failed refresh, then `/login`
- [x] Manual QA: block `/auth/token/refresh/` in DevTools → expect errors but stay signed in

Results recorded against commits `8c0fb1dfe` and `2ffbcea86` on branch `feature/MED-397-expired-session-logout`.

Ticket: [MED-397](https://linear.app/mediajira/issue/MED-397/expired-session-never-logs-the-user-out-and-retries-forever)


quanwangniuniu (approved):
Good job

PR #816, opened 2026-09-17
fix(tenant): restore caller schema after temporary tenant switches (MED-402)

FFFLAREEE:
## Summary
- Temporary tenant schema hops during org create, join, register, and SSO now go through `tenant_schema_context`, which restores the caller’s search path instead of resetting to `public`.
- `provision_tenant_schema` / `_create_tenant_tables` uses the same helper, so `Organization.save()` no longer clobbers the middleware-selected tenant mid-request.
- Adds regression tests that pin search_path after nested switches, including error paths and the create/join organization views.

## Test plan
- [x] `pytest core/tests/test_tenant_context.py` (5 passed)
- [x] `pytest authentication/tests/test_sso.py authentication/tests/test_register.py core/tests/test_onboarding.py` (48 passed)
- [ ] CI green against `prod-preview`

Made with [Cursor](https://cursor.com)


quanwangniuniu (approved):
Good job

PR #817, opened 2026-09-17
fix(spreadsheet): let keyboard focus leave the grid via Esc, Shift+Tab and skip link + MED-311

zizhenliu0427:
## Summary

Fixes [MED-311](https://linear.app/mediajira/issue/MED-311/spreadsheet-fix-keyboard-trap-in-spreadsheet-grid) — the spreadsheet grid traps keyboard focus. Tab cycles within the grid forever, and Esc only clears editing state without moving focus out. Users cannot reach the toolbar without a mouse, which violates WCAG 2.1.2 (No Keyboard Trap).

- **Tab removed from the printable-character branch** — `handleKeyDown` previously treated `Tab` as a typed character, calling `e.preventDefault()` and inserting a tab into the cell. Tab is now excluded from that branch so the dedicated Tab handler runs instead.
- **Shift+Tab boundary at first cell** — when `Shift+Tab` is pressed on cell (0,0) or with no active cell, `e.preventDefault()` is not called. The browser's natural focus order moves focus out of the grid to the preceding "Skip to toolbar" link. Shift+Tab elsewhere moves one cell left, wrapping to the end of the row above.
- **Forward Tab boundary at last cell** — Tab at the final cell also does not call `e.preventDefault()`, allowing focus to leave the grid forward.
- **Esc focuses the toolbar** — pressing `Esc` in navigation mode (not editing a cell) clears the active cell and moves focus to the first enabled toolbar button via `focusToolbar()`. If a header, sort, or export menu is open, the Esc is left to those menus' own document listeners; the grid does not also exit. In edit mode, the first Esc cancels editing; a second Esc exits the grid.
- **Skip link** — a visually-hidden "Skip to toolbar" link is rendered between the modals and the grid container. It becomes visible on focus (`sr-only focus:not-sr-only`) and uses `onClick` with `e.preventDefault()` plus `focusToolbar()`, because a plain hash anchor does not reliably move keyboard focus.
- **Toolbar semantics** — the toolbar `div` gains `ref={toolbarRef}`, `id="spreadsheet-toolbar"`, `role="toolbar"`, and `aria-label="Spreadsheet toolbar"`.
- **Grid container** — `data-testid="spreadsheet-grid"`, `aria-keyshortcuts="Escape"`, plus `role="region"` and `aria-label` so the landmark is named without using `role="grid"` (the container holds two tables, spacers and overlays, so `role="grid"` would fail axe `aria-required-children`).

## Files Changed

- `frontend/src/components/spreadsheets/SpreadsheetGrid.tsx` — `toolbarRef` / `focusToolbar()`; Tab/Shift+Tab boundaries; Esc focuses toolbar; skip link; toolbar and grid ARIA
- `frontend/e2e/spreadsheets/spreadsheet-keyboard-escape.spec.ts` — new Playwright spec (Esc exits grid; Shift+Tab plus skip link reaches toolbar)
- `frontend/e2e/spreadsheets/spreadsheet-helpers.ts` — exported `getOrCreateSpreadsheetSlug` using slug routes

## Expected Result

- `Esc` in the grid (not editing) moves focus to the first enabled toolbar button
- `Shift+Tab` from A1 moves focus to the skip link; Enter on the skip link reaches the toolbar
- `Tab` from the last cell lets focus leave the grid forward
- Tab no longer inserts a literal tab character into cells
- Existing grid navigation (arrows, Enter to edit, Esc to cancel editing) is unchanged

## Follow-up

- **Tab while editing** is out of scope (`handleInputKeyDown` does not handle Tab; no full Excel keyboard parity)
- Stale E2E helpers (`navigateToSpreadsheetAndSelectProject` still opens `/spreadsheet`, which 404s) are not fixed here to keep scope to the keyboard trap

## Test plan

- [x] `spreadsheet-keyboard-escape.spec.ts` — 2 passed
- [x] `SpreadsheetGrid.test.tsx` — 25/25 passed
- [x] `tsc` — no new errors
- [x] `next lint` on changed E2E files — clean
- [x] Manual QA: grid then Esc then toolbar focused; A1 then Shift+Tab then skip link then Enter then toolbar
- [ ] PR CI on `prod-preview`

Ticket: [MED-311](https://linear.app/mediajira/issue/MED-311/spreadsheet-fix-keyboard-trap-in-spreadsheet-grid)


quanwangniuniu (approved):
Good job！

PR #818, opened 2026-09-17
ci(cache): cache npm, pip and Playwright deps; scope buildx cache per branch + MED-322

zizhenliu0427:
## Summary

Completes MED-322 — pip and npm caching in CI. The starting point turned out to be different from the ticket's description, so the scope shifted to where the caching was actually broken.

The repository's GitHub Actions cache was sitting at **10.31 GB against a 10 GB quota across 89 entries**, so every cache in it was being evicted continuously. **79 of those 89 entries, totalling 9.81 GB, are buildkit layer blobs** exported by `lighthouse-ci.yml`'s `cache-to: type=gha,mode=max`. Adding dependency caches on top of a saturated quota would mostly have evicted them, so freeing the quota comes first and the dependency caching follows.

- **Lighthouse image cache moved off the Actions cache** — both image builds now export their layer cache to GHCR instead. GHCR storage is free and unlimited for public repositories and does not count against the 10 GB, which hands that quota back to the dependency caches below. Adds `packages: write` and a GHCR login to the job.
- **Buildx cache tags are scoped per branch in both pipelines** — every branch previously read and wrote a single shared `:cache` tag and overwrote the others, so a branch never reused its own previous build. Reads fall back branch → `cache-prod-preview` → `cache-main` → the legacy shared tag. Measured locally at **0 of 4 cache hits before the change and a hit on every repeat build after** (details under Test Plan).
- **`actions/cache@v4` on `frontend/node_modules`, skip `npm ci` on an exact lockfile hit** — this is the warmup lever. `setup-node` `cache: npm` only warms `~/.npm`; `npm ci` still unpacked ~2400 packages every run, so a cache hit cut **129s → 123s (5%)**. Caching `node_modules` itself and skipping the install on an exact `package-lock.json` hash hit cut **129s → 7s restore + 0s install (95%)**, which clears the ticket's 30% warmup target. Same key as `chromatic.yml` so the two jobs share one entry. No `restore-keys` on this entry: a partial restore of a different lockfile would be wiped by `npm ci` anyway. Miss path still has `~/.npm` as fallback and still runs `npm ci`.
- **pip caching in `code_scan`** — `setup-python` moves to v5 and gains `cache: pip` keyed on `devops/requirements-scan.txt`. `sbom_release` also moves to `setup-python@v5`, which actionlint reported as too old to run on GitHub Actions. Runner-side pip cannot be keyed on `backend/requirements.txt`: that file is never pip-installed on a runner, only inside `backend/Dockerfile`, so it is covered by the buildx layer cache.
- **Scan tooling is pinned** — bandit and pip-audit were installed unpinned, which let the scan drift between runs and left the pip cache nothing stable to key on. They now live in `devops/requirements-scan.txt` at the versions the last green run resolved.
- **Playwright browsers are cached** — keyed on the version read out of `package-lock.json`, so the key is available before `node_modules` exists. On an exact-key hit only the apt-level OS libraries are installed, since those are not part of the cached directory. Combined with the `node_modules` skip, the install step dropped **103s → 14s (86%)**.

Branch names reach the cache-scope scripts through `env` rather than being interpolated into the shell, and are sanitised to characters legal in a Docker tag. `cache-to` carries `ignore-error=true` throughout, so a cache export failure can never fail a build.

Two of the ticket's stated scope items were already in place and needed no change: `lighthouse-ci.yml` has cached `~/.npm` via `setup-node` since it was written, as has the budget Playwright step in `application_test`. A third is not implementable as written — `backend/requirements.txt` is never pip-installed on a runner.

## Files Changed

(3 files)

**CI**
- `.github/workflows/mediajira-ci.yml` — `actions/cache@v4` on `frontend/node_modules` (skip `npm ci` on hit) in `code_scan` and the Playwright step, pip cache in `code_scan`, Playwright browser cache in `application_test`, branch-scoped buildx cache tags with trunk fallbacks in `image_scan` and `application_test`, `setup-python@v5`
- `.github/workflows/lighthouse-ci.yml` — same `node_modules` cache + skip `npm ci` on hit, layer cache moved from `type=gha` to GHCR, branch-scoped tags with trunk fallbacks, GHCR login and `packages: write`

**Dependencies**
- `devops/requirements-scan.txt` — new; pins bandit and pip-audit and acts as the pip cache key

## Expected Result

- **Warmup (the ticket target) is met.** `npm ci` in `code_scan` dropped 129s → 7s (95%). Playwright install dropped 103s → 14s (86%). Both are well above 30%.
- pip scan-tool install is resolver-bound and only moved 10s → 8s. It is not the 30% lever; the npm skip is.
- A branch reuses its own image layer cache across consecutive runs instead of rebuilding cold, and a newly created branch starts warm from `prod-preview`.
- The security scan produces the same result run to run, because the tool versions no longer float.
- A cache miss still builds correctly. Missing fallback tags are logged and ignored rather than failing the build, which matters because `cache-prod-preview` and `cache-main` do not exist until this merges.
- Actions cache usage is still over the 10 GB quota until the stranded `buildkit-blob-*` entries are purged after merge (see Test Plan). New Lighthouse runs no longer write those blobs.

**Overall pipeline wall time is not the acceptance criterion and is not expected to move much** — `application_test` is still ~95 minutes, dominated by pytest. See Follow-up if a later ticket wants 30% off the whole pipeline.

## Measured on CI (two consecutive PR runs)

Baseline cold: [run 35245470404](https://github.com/quanwangniuniu/zmarkio/actions/runs/35245470404) (`059bcdced`, `~/.npm` only).
Warm after `node_modules` skip: [run 35267952137](https://github.com/quanwangniuniu/zmarkio/actions/runs/35267952137) (`2ba02a13a`).

| Step | Cold | Warm | Change |
|---|---|---|---|
| `code_scan` `npm ci` | 129s | **0s** (7s cache restore) | **−95%** |
| `code_scan` pip scan tools | 10s | 8s | −20% |
| Playwright install | 103s | **14s** | **−86%** |
| `image_scan` frontend image | 726s | 174s | −76% |
| `application_test` frontend image | 466s | 470s / 172s (see note) | variable |
| Lighthouse job | 3077s | 959s | −69% (mostly image cache) |

`~/.npm` hit alone, before the `node_modules` skip ([run 35266341258](https://github.com/quanwangniuniu/zmarkio/actions/runs/35266341258)): `npm ci` 129s → 123s. That is why skip-on-hit is required.

`application_test` frontend image was 172s on the intermediate warm run and 470s on the next — branch-scoped registry cache is working (`image_scan` 726s → 174s is stable) but that one job still misses sometimes. Lighthouse's first run after the skip change still paid 127s for `npm ci` (raced the first `node_modules` save); later runs should skip.

## Follow-up Recommendation

The 30% **warmup** criterion is met above. Cutting **overall** wall time 30% is a different ticket: the critical path is `application_test` at ~95 minutes, of which a single `pytest -v` step is ~62 minutes.

The cause is throughput rather than any slow test. That step runs thousands of tests across workers `gw0` and `gw1` — `-n auto` resolves to two, because `ubuntu-latest` is a two-core runner. Two of the slowest entries are per-worker test-database setup, which is the migration history being replayed once per worker.

**Running the suite across more than one runner is what moves wall time.** Three forms, in decreasing order of how far they sit outside this ticket:

- **Self-hosted runners** are named in the ticket's Out of Scope and are not proposed.
- **A larger GitHub-hosted runner** is the cheapest change available — one `runs-on:` line, no change to job topology, no additional billed minutes, and free of charge for public repositories. It is not what the Out of Scope line excludes, since larger runners are still GitHub-hosted, but an organisation administrator has to provision one before the label resolves.
- **Matrix sharding** across standard runners changes only `.github/workflows/mediajira-ci.yml`, needs no infrastructure from anyone, and respects the self-hosted exclusion. Its billed-minute cost depends on this PR landing first, because the duplicated per-shard cost is almost entirely the frontend image build — which is what the branch-scoped registry cache serves.

`image_scan` and `application_test` build the same two images independently. Since `image_scan` is not on the critical path this costs no wall-clock time, but it is billed runner time per run. Building once and passing the image between jobs would recover it.

The legacy `:cache` entry in each `cache-from` list is transitional. It keeps branches that are already in flight warm during the switchover and should be removed once `prod-preview` has built once on the scoped tags.

Two smaller items were left alone deliberately. `sbom_release` still runs `pip install cyclonedx-bom` unpinned with no cache; it has the same drift problem the scan tools had, but it only fires on tags, so there is no warm-path time to win and the value would be reproducibility only. Separately, `mediajira-ci.yml` uses `setup-node@v4` while `lighthouse-ci.yml` uses v6, and the two action versions use different `~/.npm` cache key prefixes. The `node_modules` entry now uses one shared key with Chromatic.

## Test Plan

The buildx cache change was A/B tested locally before pushing, against a local registry using a synthetic image that mirrors `backend/Dockerfile`'s shape — a slow cacheable dependency layer followed by a source layer. **Local buildkit state was wiped before every build**, so a fresh CI runner is simulated and only the registry cache can produce a hit. An earlier attempt without that wipe produced false hits and was discarded.

- [x] **Cache thrash reproduced on the current behaviour** — two branches with different dependency sets alternating against one shared `:cache` tag: **0 of 4 cache hits**, every build cold at 25–26s, including each branch's second run with its dependencies unchanged
- [x] **Branch isolation confirmed after the change** — same sequence with branch-scoped tags: each branch's second run **hit** and dropped from 25s to 5–6s, an 80% reduction, despite the other branch building in between
- [x] **Fallback chain confirmed** — a brand-new branch whose dependencies match `prod-preview` **hit** off `cache-prod-preview` at 6s versus 25s cold
- [x] **Missing fallback refs are non-fatal** — `cache-main` and the legacy tag did not exist; builds logged `not found` and returned rc=0
- [x] **Concurrent writes to one scoped tag are safe** — two parallel builds both writing `cache-feature-MED-322-ci-caching`, mirroring `image_scan` and `application_test`, both succeeded with no exporter errors, and a subsequent fresh build still hit
- [x] **`~/.npm` alone is not enough** — CI hit the npm download cache and still ran `npm ci` in 123s vs 129s cold (5%)
- [x] **`node_modules` skip meets the 30% warmup target** — `code_scan` install 129s → 0s (7s restore); Playwright 103s → 14s. Cold path still succeeded on the first run of each revision
- [x] **pip cache measured** — `pip install -r devops/requirements-scan.txt`: 10s cold, 8s warm on CI. Modest because most of that time is resolver work rather than download
- [x] **Pinned versions resolve** — `bandit==1.9.4` and `pip-audit==2.10.1` install cleanly
- [x] **actionlint, diffed against `prod-preview`** — 9 findings before, 7 after, **none introduced**. The two resolved are both `setup-python@v4` runner-too-old. The 7 remaining are pre-existing and unrelated: untrusted `head.ref` in `block`, SC2086 in the Bandit and Trivy steps, SC2034 in the Lighthouse wait loops
- [x] **Both workflows parse, and every `cache-from` / `cache-to` line asserted to be a well-formed buildx spec** — an earlier revision had explanatory comments inside the `cache-from` block scalar, where buildx would have received them as cache specs
- [x] **Tag sanitiser** — exercised against real branch names including `feat/MED-296-eliminate-task-hierarchy-n+1` and a shell-metacharacter string; all produce valid Docker tags
- [x] **Cache hit-rate validated on consecutive CI runs** — `~/.npm` + pip + Playwright + `node_modules` all hit on the follow-up run. Quota is still above 10 GB because old `buildkit-blob-*` entries remain; they are no longer being written

After this merges and `prod-preview` has built once, the stranded buildkit blobs are worth purging:

```
gh api "repos/{owner}/{repo}/actions/caches?per_page=100" --jq '.actions_caches[] | select(.key|startswith("buildkit-blob")) | .id' | xargs -I{} gh api -X DELETE "repos/{owner}/{repo}/actions/caches/{}"
```

Ticket: [MED-322](https://linear.app/mediajira/issue/MED-322/devops-speed-up-ci-with-pip-and-npm-caching)


quanwangniuniu (approved):
Good job

PR #820, opened 2026-09-18
MED-358: Replace the full-table slug scan on the generate path

oolonglingt:
Removed the full-table slug scan, moved slug uniqueness enforcement to the database with conflict retries, and added concurrent batch test coverage.


quanwangniuniu (approved):
Good job

PR #823, opened 2026-09-19
MED-447 refactor(ci): cut PR feedback from ~90 min to ~35 min by building images once and splitting application_test

oliverwu1024:
## Problem

Every PR waited 85–102 min for `application_test`, plus a separate 22–27 min Lighthouse run, whatever the size of the change. (The ticket quotes 165.5 min, from a run on 3 Sep; by 14–17 Sep it was 85–102 min.)

Measuring per-step times showed the backend pytest step was ~75% of the job (62–72 min), while image builds were only 4–11 min. So sharing images — the ticket's first suggestion — was worth at most ~10 min; splitting the test suite was what mattered.

## What this does

- **Build once.** New `build_images` job builds the backend, frontend and database images and passes them to every job as artifacts. Previously each of `application_test`, `image_scan` and Lighthouse built them, and `docker compose up` then rebuilt them a fourth time, so the images those jobs built were never actually used. Test jobs now start compose with `--no-build`.
- **`application_test` split into parallel jobs:** 4 backend shards, serial tests, frontend tests, studio tests, budget E2E and Lighthouse. A job still named `application_test` aggregates them, so the required check, the `deploy` gate and the Discord notification are unchanged.
- **Shards split by measured run time.** `pytest_collection_modifyitems` in `backend/conftest.py` (active only when `PYTEST_SHARD_INDEX`/`PYTEST_SHARD_COUNT` are set) assigns whole modules using `backend/test_durations.json`. Balancing by test count gave 12–41 min shards; by time they are 20–24 min. There is no hand-kept path list, and `backend_serial_tests` fails the build unless the shards exactly partition the collected suite.
- **Coverage gate preserved.** Shards run with `--cov-fail-under=0` and upload their data; `backend_coverage` combines it and enforces 85%.
- **Lighthouse moved into this pipeline** and reuses the shared images; `lighthouse-ci.yml` is deleted. It still reports as its own `lighthouse` check and, as before, does not gate `deploy`.
- **4 test workers** (`-n logical`; `-n auto` counted physical cores and used 2 of the runner's 4 vCPUs).
- **Fast password hasher in tests** (test-only; ~931 hashing call sites, no test asserts the algorithm).
- **`-o timeout_func_only=true`** in shards, plus `timeout-minutes: 45`. With 4 workers, per-worker test-DB setup takes 530–556 s against the 600 s `--timeout`, and once exceeded it.

## Bug found and fixed on the way

`backend/backend/ci_settings.py` did `from .settings import *`, but `settings.py` ends with `__all__ = ('celery_app',)`, so CI saw **zero installed apps**. The "Run migration checks" step therefore never checked or applied anything (`Apply all migrations: (none)`), despite `CICD_README.md:74` documenting it as the safeguard against forgetting a migration. The backend tests would not reliably have caught it either: in tests, tenant tables are created directly from current models by `provision_tenant_schema()`, not from migration files.

Verified both ways: with a model field added and no migration the check now **fails** and names it; after reverting, it passes. No migrations are missing today.

**Reviewers should know:** from now on this check really does fail PRs that change a model without a migration.

## Results

| | Before | After |
|---|---|---|
| PR feedback (wall) | 85–102 min | **33–42 min** |
| Backend tests | one step, 62–72 min | 4 shards, 20–24 min each |
| Lighthouse | 22–27 min | 13–15 min |
| Image builds per PR | 3 (+1 by compose) | 1 |
| Total runner minutes | ~127 | **~160–175** |

**Trade-off:** wall-clock feedback drops ~60%, but total runner minutes rise ~26%, because each parallel job boots its own stack. The duplicated image building the ticket flagged is gone; the extra cost is the parallelism itself. Fewer shards would trade feedback time back for runner minutes.

**Parity:** 6,495 backend tests (6,471 passed + 24 skipped) plus 93 serial, combined coverage 92% (15,047/1,172), frontend 1,181 with the `test:ci` gate, budget E2E 6 Playwright tests — all matching `prod-preview`.

## Note on MED-322

MED-322 (#818) merged while this was in QA and changed the same files. Its work is carried over, not dropped:

- Buildx cache scoping and the GHCR login now live in `build_images`, the only job that still builds images (hence one "Compute cache scope" step instead of two).
- Its Playwright and `node_modules` caches stay in the budget E2E job.
- Its npm/pip caching in `code_scan` and `setup-python@v5` are untouched.
- `lighthouse-ci.yml` stays deleted: its MED-322 changes fixed the image builds that job no longer performs.

`build_images` took 17.3 min in the verification run, the first write to the new per-branch cache tag, so it built close to cold. Later runs on a branch should be warmer, but that is unconfirmed from one run.

## Verification

Seven CI runs on the branch via a temporary trigger, added and removed (the net diff contains no trigger):

| Run | Result | Wall | Note |
|---|---|---|---|
| 35204736524 | E2E failed | 58.1 min | exposed the `ci_settings` bug; shards unbalanced |
| 35210564481 | passed | 32.7 min | time-based balancing |
| 35213721148 | passed | 48.4 min | migration fix; build-cache miss |
| 35218580787 | shard 1 failed | 40.1 min | DB setup hit the 600 s timeout |
| 35224323405 | passed | 34.1 min | timeout fix |
| 35335829053 | passed | 38.5 min | after merging prod-preview |
| 35417005963 | **passed** | **42.3 min** | final, after the MED-322 merge |

## Follow-ups (separate tickets)

1. `frontend/Dockerfile`: `RUN chown -R` costs ~175 s; `COPY --chown` would remove that layer.
2. `jlumbroso/free-disk-space` varies 1–9 min between runners and often sets the critical path.
3. Per-test tenant schemas: every `Organization` a test creates runs `CREATE SCHEMA` plus ~220 `CREATE TABLE` (`core/services/tenant.py`), the main reason the median backend test costs 1.39 s. Worth more than further CI parallelism.
4. Three tests each burn ~120 s, likely unmocked network: `automationWorkflow … test_list_workflows`, `google_ads … test_delete_ad_success`, `notion_editor … test_revision_snapshot_accuracy`.
5. A lighter CI tier for docs-only PRs (the ticket's third suggestion), deferred because it needs team agreement on which paths are safe to skip.
6. `backend/test_durations.json` drifts as tests change; shards stay correct but less even, so regenerate occasionally.

Ticket: MED-447


quanwangniuniu (changes requested):
Why we removed lighthouse?

backend/backend/ci_settings.py
quanwangniuniu: how does this change even work?

backend/test_durations.json
quanwangniuniu: Are you sure this could have an effect on the real ci pipeline?

quanwangniuniu (changes requested):

PR #826, opened 2026-09-21
(MED-248) fix(meta-ads): namespace creative preview cache by account 

FFFLAREEE:
## Summary
- Scope Meta creative preview cache keys by `meta_creative_id` + ad account so cross-account platform IDs cannot collide, with preview fetch moved into a thin service layer and covered by pytest.
- Clear frontend preview state when the selected ad account changes (Creatives / Ads hierarchy) and reset VideoModal media on close to avoid stale preview UI.
- Add a CI-safe mocked Playwright project (`meta-ads-mock`) that verifies account A → B switch does not reopen account A preview markers.

## Test plan
- [ ] `docker compose -p mediajira-v2 -f docker-compose.dev.yml exec backend pytest backend/utils/tests/test_ad_preview_cache_keys.py backend/meta_ads/test_creative_preview_cache.py`
- [ ] `docker compose -p mediajira-v2 -f docker-compose.dev.yml exec -e E2E_USE_EXISTING_SERVER=1 -e BASE_URL=http://nginx:80 frontend npx playwright test --project=meta-ads-mock --retries=0`
- [ ] Manual: open Meta Ads creatives preview on account A, close, switch to account B, reopen preview — confirm B content (not A).
- [ ] Confirm Google Ads is out of scope for this ticket.

Made with [Cursor](https://cursor.com)


quanwangniuniu (changes requested):
Here are 7 issues I’d block on before merging this PR, with the reason each one has to be fixed.

1. The new cache helper is imported from `utils.ad_preview_cache_keys`, but that helper is not present in the repo tree I inspected, and the import would fail at runtime if it isn’t added in the same PR.
- Why fix it: this is a hard runtime break. The service will crash on import before any preview logic runs, which makes the feature unusable.
- Risk: import-time failure, production outage in the preview endpoint, and CI failures if the file isn’t actually included.

2. The cache key scheme is still not fully invalidation-safe.
- Why fix it: caching by `platform + account_id + creative_id + variant` helps with cross-account collisions, but it does not protect against stale preview data after the creative is edited, re-linked to a different ad, or its asset metadata changes.
- Risk: users can see old iframe HTML or stale thumbnails even after the underlying creative changed, which is a correctness bug and a customer-facing trust issue.

3. The preview lookup trusts the first ad returned by `creative.ads.order_by("-updated_at").first()`.
- Why fix it: a creative can legitimately be linked to multiple ads over time, and the most recently updated ad may not be the one the user is currently viewing or expecting.
- Risk: preview can render for the wrong ad or wrong account, especially during sync races or after a creative is re-used or migrated. That creates wrong preview data and cross-account confusion.

4. The code silently tolerates a missing/empty iframe source instead of failing the request.
- Why fix it: `re.search(r'src="([^"]+)"', body)` can fail, and when it does, `iframe_src` becomes an empty string while the function still returns a valid-looking payload.
- Risk: the API responds with a “successful” payload that cannot actually render a preview. That produces broken UI and makes the error impossible to diagnose from the backend response.

5. The HTML parsing is too brittle for Meta’s real payloads.
- Why fix it: Meta preview HTML is not guaranteed to use exactly `src="..."` with double quotes. It can also come back in other forms, with extra attributes, or encoded markup.
- Risk: valid previews will randomly fail if the response format changes even slightly. This is a classic “works in one payload, breaks in production” bug.

6. The cache bypasses any validation of the payload before returning it.
- Why fix it: the code does `cache.get(key)` and returns the object immediately, with no check that it still matches the current creative, account, or ad format. If the cache is stale or poisoned, there is no safe guard.
- Risk: stale or mismatched preview content can be served to the wrong user or the wrong selection state. This is especially dangerous in multi-account interfaces.

7. There is no invalidation path when the account context changes or the preview is closed and reopened.
- Why fix it: the PR adds account-scoped cache keys, but the app still has to ensure old preview state is not retained in the frontend or reused across account switches. If the backend cache is reused across sessions or stale UI state remains, you still get wrong content.
- Risk: users can switch from account A to account B and accidentally still see A’s preview, which is exactly the kind of bug that triggers support tickets and compliance concerns.

FFFLAREEE:
> Here are 7 issues I’d block on before merging this PR, with the reason each one has to be fixed.
> 
> 1. The new cache helper is imported from `utils.ad_preview_cache_keys`, but that helper is not present in the repo tree I inspected, and the import would fail at runtime if it isn’t added in the same PR.
> 
> * Why fix it: this is a hard runtime break. The service will crash on import before any preview logic runs, which makes the feature unusable.
> * Risk: import-time failure, production outage in the preview endpoint, and CI failures if the file isn’t actually included.
> 
> 2. The cache key scheme is still not fully invalidation-safe.
> 
> * Why fix it: caching by `platform + account_id + creative_id + variant` helps with cross-account collisions, but it does not protect against stale preview data after the creative is edited, re-linked to a different ad, or its asset metadata changes.
> * Risk: users can see old iframe HTML or stale thumbnails even after the underlying creative changed, which is a correctness bug and a customer-facing trust issue.
> 
> 3. The preview lookup trusts the first ad returned by `creative.ads.order_by("-updated_at").first()`.
> 
> * Why fix it: a creative can legitimately be linked to multiple ads over time, and the most recently updated ad may not be the one the user is currently viewing or expecting.
> * Risk: preview can render for the wrong ad or wrong account, especially during sync races or after a creative is re-used or migrated. That creates wrong preview data and cross-account confusion.
> 
> 4. The code silently tolerates a missing/empty iframe source instead of failing the request.
> 
> * Why fix it: `re.search(r'src="([^"]+)"', body)` can fail, and when it does, `iframe_src` becomes an empty string while the function still returns a valid-looking payload.
> * Risk: the API responds with a “successful” payload that cannot actually render a preview. That produces broken UI and makes the error impossible to diagnose from the backend response.
> 
> 5. The HTML parsing is too brittle for Meta’s real payloads.
> 
> * Why fix it: Meta preview HTML is not guaranteed to use exactly `src="..."` with double quotes. It can also come back in other forms, with extra attributes, or encoded markup.
> * Risk: valid previews will randomly fail if the response format changes even slightly. This is a classic “works in one payload, breaks in production” bug.
> 
> 6. The cache bypasses any validation of the payload before returning it.
> 
> * Why fix it: the code does `cache.get(key)` and returns the object immediately, with no check that it still matches the current creative, account, or ad format. If the cache is stale or poisoned, there is no safe guard.
> * Risk: stale or mismatched preview content can be served to the wrong user or the wrong selection state. This is especially dangerous in multi-account interfaces.
> 
> 7. There is no invalidation path when the account context changes or the preview is closed and reopened.
> 
> * Why fix it: the PR adds account-scoped cache keys, but the app still has to ensure old preview state is not retained in the frontend or reused across account switches. If the backend cache is reused across sessions or stale UI state remains, you still get wrong content.
> * Risk: users can switch from account A to account B and accidentally still see A’s preview, which is exactly the kind of bug that triggers support tickets and compliance concerns.

Thanks for the detailed review. I checked each point against the current branch and the scope of MED-248: namespace creative preview cache keys by ad account so identical platform creative IDs cannot collide across accounts.

### Changes made following this review

* An empty or missing `iframe_src` now returns `502` with code `missing_iframe_src`, and the invalid payload is not cached.
* The `src` parser now accepts single or double quotes and optional whitespace around `=`.

### Already included in the PR

* `backend/utils/ad_preview_cache_keys.py` and its tests.
* Frontend preview state is cleared when `adAccountId` changes.
* `VideoModal` clears its in-memory preview when closed.
* Playwright covers the user-visible A → B account-switch flow without leftover A content.

### Follow-up areas outside the stated MED-248 acceptance criteria

* Edit/sync-time cache invalidation.
* Selecting a specific linked ad from the client.
* Additional validation of cached payload contents.

---

### 1. Missing `utils.ad_preview_cache_keys`

The helper is included in this PR at:

* `backend/utils/ad_preview_cache_keys.py`
* `backend/utils/tests/test_ad_preview_cache_keys.py`

`get_creative_preview` imports this helper, and pytest covers key construction and cross-account isolation.

If the file was missing from the inspected tree, please pull the latest version of `fix/MED-248-namespace-creative-preview-cache-keys`. The import succeeds on the current branch.

### 2. Cache key is not fully invalidation-safe

Agreed: a key containing `platform + account_id + creative_id + variant` does not invalidate immediately when a creative is edited, re-linked, or updated during sync.

MED-248 addresses cross-account key collisions. Edit/sync-time invalidation is not part of the stated acceptance criteria and requires a separate lifecycle decision, such as:

* including a version or update timestamp in the key; or
* deleting relevant keys when creative data is updated.

Entries currently expire after one hour through:

```python
CREATIVE_PREVIEW_CACHE_TTL_SECONDS = 3600
```

The ticket explicitly excludes server-side TTL tuning. I suggest tracking edit/sync-time invalidation as a follow-up rather than making it a blocker for the account namespacing change.

### 3. Preview uses `creative.ads.order_by("-updated_at").first()`

Meta’s preview endpoint is `/{ad_id}/previews`, so the service must select a linked ad.

The current frontend request only supplies the creative ID:

```text
GET /creatives/{id}/video_source/
```

Even when the preview is opened from the campaign hierarchy, the frontend passes `ad.creative.id` and does not send the selected ad ID.

The “latest linked ad by `updated_at`” rule already existed in the view. This PR moved that logic into the service while keeping its existing behavior.

Previewing the exact ad selected by the user would require an API and UI change: the client would need to pass a `meta_ad_id`, and the backend would need to validate that it belongs to the requested creative and accessible account. I suggest handling that separately from MED-248.

### 4. Empty `iframe_src` returned 200

Fixed.

`VideoModal` renders `iframe_src`; it does not render `iframe_html`. Previously, an empty `iframe_src` returned `200`, displayed “No preview data available,” and was cached for one hour.

The service now fails before `cache.set()` when the extracted source is missing or blank:

* status: `502`
* code: `missing_iframe_src`
* invalid payload: not cached

Tests now verify that:

* the service raises `CreativePreviewError`;
* the API view returns `502` with `missing_iframe_src`;
* the failed result is absent from cache; and
* a later valid Graph response succeeds, confirming that the failure did not poison the cache.

### 5. HTML parsing was too brittle

Hardened for the formats raised in the review.

The parser now accepts single or double quotes and optional whitespace around `=`:

```python
src\s*=\s*["']([^"']+)["']
```

This supports:

```html
src="..."
src='...'
src = "..."
src = '...'
```

Meta’s preview body is expected to contain a single iframe. If no usable source is found, the service now fails closed with the `502` described in point 4.

A full DOM parser and explicit iframe-element selection would be a broader hardening change. If real Meta payloads demonstrate that this is needed, it can be handled separately.

### 6. Cache hit returns the payload without re-validation

The cache key is now:

```text
ad_creative_preview:meta:{account_id}:{meta_creative_id}:{ad_format}
```

It binds the entry to the platform, account, platform creative ID, and preview format. This removes the cross-account collision targeted by MED-248.

Re-validating fields such as `creative_id` and `ad_format` inside the cached payload could provide additional defence against malformed or incorrectly written cache entries. However, it would not detect stale preview HTML after an edit because the account, creative ID, and format would still match. That belongs to the edit/sync-time invalidation concern in point 2.

The cache is intentionally shared by users who are authorised to access the same ad account. Authorisation still occurs before the service call through:

```python
ad_account_id__in=_accessible_ad_accounts_for_user(request.user)
```

### 7. No invalidation when the account changes or the modal closes

The frontend state handling is already implemented:

* `CreativesPanel` and `CampaignHierarchyTable` clear `previewCreativeId` and the preview title when `adAccountId` changes.
* `VideoModal` clears its in-memory preview when it is closed or has no creative ID, preventing the previous iframe from flashing when reopened.

The Playwright test covers the user flow supported by the full-screen modal:

```text
open account A preview
→ close the modal
→ switch to account B
→ reopen the preview
→ show the B marker only
```

It confirms that the A marker and preview controls are absent after switching and that reopening shows only B’s content.

The component code enforces the account-change reset. The E2E covers the visible close/switch/reopen workflow. Backend entries remain separated by account until their TTL expires, so account B cannot reuse account A’s cache key.

quanwangniuniu (approved):
Good fix@!

PR #827, opened 2026-09-21
MED-274: Custom KPI builder with formula editor

zizhenliu0427:
## Summary

Implements MED-274 — custom KPI builder with formula editor. Users can define project-scoped KPIs with formulas (e.g. `revenue / spend`, `clicks / impressions * 100`) that evaluate against the project's Meta Insights warehouse data. The formulas reuse the existing spreadsheet formula engine via a new identifier resolver hook.

- **Formula engine extension** — `formula_engine.py` gains an optional `identifier_resolver` callback and underscore-aware tokenizer, so metric names like `landing_page_views` resolve as variables instead of raising `#REF!`. The hook is opt-in: every new branch is gated on a resolver being passed, so sheets — which pass none — behave exactly as before, confirmed by the existing engine tests still passing unchanged.
- **KPI registry** — new `kpi_registry.py` provides formula validation (length, token-count and nesting-depth limits, cell-reference rejection, unknown metric/function detection, plus a probe parse that rejects structurally broken formulas like `revenue /` before they can be saved), metric resolution from `MetaInsightDaily`, and evaluation. The sandbox is structural rather than a denylist: the engine is a hand-written recursive-descent parser with no `eval`/`exec`, so a formula can only ever produce a number, string, boolean or error code. A `#NODATA` error code distinguishes empty warehouse data from genuine formula faults.
- **Backend CRUD + preview** — `CustomKPI` model with project FK and unique name constraint, four serializers, four views (list/create, detail, preview, metric catalog), migration 0004.
- **Frontend builder UI** — `CustomKPIPanel` on the Overview dashboard with `KPIBuilderDialog` (create/edit) and `FormulaEditor` (CodeMirror 6 with metric autocomplete, single-line enforcement, inline error display, 400ms debounced live preview).

## Files Changed

(28 files; key areas)

- **`backend/spreadsheet/formula_engine.py`** — `IdentifierResolver` hook + underscore-aware tokenizer
- **`backend/report/`** — new `kpi_registry.py` (metric catalog, formula sandbox, evaluation); `CustomKPI` model + migration 0004; 4 serializers, 4 views, 4 routes
- **`backend/report/tests/`** — `test_kpi_registry.py` + `test_kpi_api.py` + `conftest.py`
- **`frontend/src/components/dashboard/kpi/`** — new `CustomKPIPanel`, `KPIBuilderDialog`, `FormulaEditor` (CodeMirror 6), `kpiErrors.ts`, `formatKPIValue.ts`
- **`frontend/src/lib/api/reportApi.ts`** + **`frontend/src/types/report.ts`** — KPI API client + types
- **`frontend/src/components/overview/OverviewContent.tsx`** — mount `<CustomKPIPanel />`
- **`frontend/src/components/ui/select.tsx`** — shared primitive: `SelectContent` z-50 → z-[10001]. It portals to `<body>`, so under the z-[10000] dialog layer every `Select` inside a `Dialog` opened behind the overlay and could not be clicked. Also fixes `SlackIntegrationModal`
- **`backend/meta_ads/management/commands/seed_meta_insights.py`** — local-only command to seed fake `MetaInsightDaily` rows, since the dev warehouse is empty and KPIs otherwise have nothing to show
- **`frontend/e2e/reporting/kpi-builder.spec.ts`** — 8 Playwright tests

**Build fixes (6 files, pre-existing, not KPI-related — see note below)**

## Note on Build Fixes

This PR includes fixes to 6 pre-existing files that are unrelated to the KPI feature. They were necessary to run the Playwright E2E tests:

> Playwright requires a working build → `npm run build` was already broken on `main` → `skip-step-real-session.spec.ts` had an orphaned code block that made tsc skip semantic analysis project-wide → removing it un-masked 14 real type errors across 6 other files → those had to be fixed too, or build still fails.

A parse error makes tsc report only syntactic diagnostics and skip semantic analysis for the whole program, which is why `next.config.mjs`'s "TSC errors are now 0" comment was false while `ignoreBuildErrors: false` appeared to be enforcing. The 6 files carrying errors are not the same as the 6 files edited below — the `tsconfig` bump cleared `campaignApi.test.ts` and `booking-link-manager.spec.ts`, and the `tasks-helpers.ts` fix cleared `task-parent-cycle.spec.ts`.

Every step was forced by the previous one. An alternative would have been setting `ignoreBuildErrors: true` in `next.config.mjs` temporarily, but fixing the root cause benefits everyone — `npm run build` now succeeds for the first time, and `tsc` reports genuinely 0 errors.

The fixes are isolated in their own commits (`258f5ee25` and `485a3dcde`) and can be split into a separate PR if the reviewer prefers:

- `e2e/agent/skip-step-real-session.spec.ts` — remove orphaned block
- `tsconfig.json` — `lib` ES2017 → ES2022 (for `.at()`)
- `e2e/tasks/tasks-helpers.ts` — add missing `summary` to `DraftTaskFixture` (was a latent runtime bug)
- `e2e/auth/oauth.spec.ts` — fix closure read
- `e2e/navigation/sidebar-navigation.spec.ts` — null guard
- `e2e/notification/preference-refetch-on-focus.spec.ts` — null guard

## Expected Result

- Users can create, edit, and delete custom KPIs scoped to a project.
- Formula errors (syntax, unknown metrics, and division by zero on a project that has data) are surfaced inline while typing via the 400ms debounced preview.
- A project with no warehouse rows shows "No data for this period" instead of a misleading `#DIV/0!`, rendered as neutral information rather than a red error, and the KPI stays saveable.
- Custom KPIs persist across page reloads and appear on the Overview dashboard.
- CodeMirror autocomplete suggests available metric names and formula functions.
- `npm run build` succeeds (first time on this branch — previously broken by pre-existing type errors).

## Follow-up Recommendation

- **Cross-platform metrics:** Only Meta (`MetaInsightDaily`) data is available. "Blended ROAS" across Google Ads + Meta is not achievable until Google Ads daily metrics are stored. Worth raising on the ticket.
- **Empty-project UX:** "No data for this period" is technically correct but could link to a data connection guide.
- **Cross-project KPI library:** Out of scope per ticket. A shared template system could be built later.
- **`task-parent-cycle.spec.ts`:** unstable independent of this work, verified by re-running against the pre-change files — test 75 is flaky (fails, passes on retry) and test 139 fails consistently both before and after. Test 186 was not isolated. Serial mode means one failure skips the rest, so the file rarely completes.

## Test Plan

- [x] `test_kpi_registry.py` + `test_kpi_api.py` — **67 passed** (formula sandbox, evaluation, metric resolution, CRUD, preview, access control, cross-project isolation)
- [x] `spreadsheet/tests/` — formula engine, services, cell sync and performance suites pass unchanged, confirming the resolver hook does not affect sheets
- [x] `kpi-builder.spec.ts` — **8/8 passed** (inline errors, error recovery, no-data, format dropdown, create, edit, delete)
- [x] Frontend lint — clean on all changed files
- [x] `tsc` — 0 errors (fixed pre-existing type errors)
- [x] `npm run build` — succeeds
- [x] Manual QA: create KPI `revenue / spend` → preview shows value or #NODATA → saves → persists on reload

Results recorded against commits `741cb3cc6` through `319fb8df2` on branch `feature/MED-274-kpi-builder`.

Ticket: [MED-274](https://linear.app/mediajira/issue/MED-274/reporting-custom-kpi-builder-with-formula-editor)


quanwangniuniu (changes requested):
1. **Percent-formatted KPIs are displayed with the wrong scale** — `frontend/src/components/dashboard/kpi/formatKPIValue.ts` (`formatKPIValue`)

   The `percent` format only appends `%` to the raw numeric value; it does not multiply the value by 100. For example, `clicks / impressions` evaluates to `0.025`, but the UI displays `0.03%` instead of the expected `2.5%`.

   This makes percentage KPIs misleading and can cause users to make decisions based on incorrect metrics. Either multiply percentage values by 100 in the formatter or clearly define that users must include `* 100` in every percentage formula and enforce that behavior consistently in the UI.

2. **Model validation can be bypassed because `CustomKPI.save()` does not call `full_clean()`** — `backend/report/models.py` (`CustomKPI.clean`)

   Formula validation is implemented in `clean()`, but Django does not invoke `full_clean()` automatically when `save()` is called. Any KPI created through the ORM, an admin action, a data migration, or another internal code path can therefore persist an empty, malformed, or unknown-metric formula.

   This breaks the model’s stated invariant that invalid formulas never reach the database. The validation should be enforced at every persistence boundary, for example by calling `full_clean()` in an overridden `save()` method or by ensuring all creation paths explicitly call model validation.

3. **The list endpoint computes warehouse data for projects the user cannot access** — `backend/report/views.py` (`CustomKPIListCreateView.get_serializer_context`)

   `get_queryset()` correctly limits KPIs to active project members, but `get_serializer_context()` independently resolves the raw `project` query parameter and calls `resolve_metric_values()` before checking whether that project belongs to the user’s accessible project set.

   An authenticated user can submit another project’s slug or ID and cause the server to query that project’s warehouse data. Although the current response may contain no KPI rows, this still bypasses the intended access boundary, adds unnecessary database work, and could become a data leak if the serializer or response behavior changes later. The project should be resolved and authorized once, then reused for both the queryset and metric snapshot.

4. **The preview request can report a stale result for the current formula** — `frontend/src/components/dashboard/kpi/KPIBuilderDialog.tsx` (`useEffect` debounce logic)

   The code protects against stale successful responses with `requestId`, but it does not invalidate an in-flight request when the dialog closes or when the formula is cleared. A previous request can therefore complete after the input has been emptied or the dialog has been closed and still update `preview`, `formulaError`, or `previewing`.

   This can show an error or value for a formula that is no longer in the editor and can leave stale validation state when the dialog is reopened. Increment or invalidate the request ID during cleanup and when the form is reset, and optionally cancel the request with an `AbortController`.

zizhenliu0427:
Thanks — all four are real, and all four are fixed in `82a35d0c9`. Details below, including two places where I made a judgement call rather than taking the suggested fix verbatim.

---

**1. Percent-formatted KPIs displayed with the wrong scale**

Fixed in `formatKPIValue.ts`: `percent` now scales by 100 before formatting, so `clicks / impressions` = `0.025` renders as `2.5%` instead of `0.03%`.

I took the "multiply in the formatter" option rather than "document that users must write `* 100`", because it matches what a spreadsheet's percent format does and because the alternative leaves the trap in place — a user picking Percent and writing the obvious ratio would still get a wrong number.

To make the convention explicit in the UI, as you asked: the builder now shows a hint under the Format select whenever Percent is chosen, telling the author to write the ratio itself and not to add `* 100`.

Added `src/__tests__/components/dashboard/kpi/formatKPIValue.test.ts` to lock the scaling down (0.025 → 2.5%, 0.5 → 50%, 1 → 100%), plus a case asserting `number` and `currency` are *not* scaled.

No stored KPIs were affected — there is no production data for this feature yet.

---

**2. Model validation bypassed because `save()` does not call `full_clean()`**

Correct, and worse than the code suggested: the class docstring claimed *"The formula is validated on save so a KPI that cannot be evaluated never reaches the database"*, which was simply false for any non-serializer path.

`CustomKPI.save()` now calls `self.full_clean()` before `super().save()`, and the docstring has been rewritten to state what actually happens. The API path is unaffected — the serializers still validate first and return a proper 400 with a field error, so `full_clean()` only ever fires as the backstop.

Tests: `test_orm_create_also_rejects_an_invalid_formula` creates through `CustomKPI.objects.create()` with an unknown metric, a malformed formula and a blank formula, asserting each raises and nothing is persisted; `test_orm_create_with_a_valid_formula_still_works` guards the happy path.

---

**3. List endpoint computes warehouse data for inaccessible projects**

Fixed. `get_queryset()` and `get_serializer_context()` no longer resolve the `project` parameter independently. Both now call `_requested_project_pk()`, which resolves the slug/id, checks `ProjectMember` once, caches the result on the request, and returns a `(pk, scoped)` pair. When the caller is not a member, `scoped` is `False`: the queryset becomes `none()` and no metric snapshot is built, so `resolve_metric_values()` is never reached.

One deliberate deviation: I kept the response as **200 with an empty list** rather than raising 403. Returning 403 would confirm the existence of a project the caller cannot see, and the existing API contract already returns an empty list for a project with no visible KPIs. The access boundary you identified — doing warehouse work for someone else's project — is closed either way.

`test_listing_an_unauthorized_project_does_not_touch_the_warehouse` patches `resolve_metric_values` and asserts `assert_not_called()`, since an empty response alone would not have proven the fix. `test_listing_an_authorized_project_still_aggregates` confirms the guard did not switch the normal path off.

---

**4. Stale preview result for the current formula**

Fixed. The debounce effect's cleanup now increments `requestIdRef` in addition to clearing the timer, so any request already in flight is invalidated on teardown. React runs that cleanup before the next effect, which covers all three cases you described: the formula being cleared, the dialog closing, and unmount.

I did not add an `AbortController`. The increment is sufficient for correctness — a late response is discarded before it can touch `preview`, `formulaError` or `previewing` — and it avoids threading cancellation through the shared axios client. Happy to add it if you would rather the request itself be cancelled rather than ignored; it is a small change.

---

**Verification** (against `82a35d0c9`)

- backend pytest — 73 passed (67 before, 6 new covering the above)
- `formatKPIValue` jest — 5 passed (new)
- `kpi-builder.spec.ts` Playwright — 9 passed
- `tsc --noEmit` — 0 errors
- `next lint` — clean on all changed files

quanwangniuniu (approved):
Good job

PR #828, opened 2026-09-21
MED-372: Define PlatformSpec and the platform registry

PianoMaksim:
Add a serializable platform contract and registry so platforms with scalar or repeated text fields can declare their rules as data. Meta is the only supported registry entry; Google RSA is included as an isolated design-check spec.

- Validate authored configuration, including field limits, cardinality, CTA policies and slug references.
- Define fields for non-Meta ad formats, including repeated headlines and descriptions in the Google RSA design-check spec.
- Provide detached presentation data without internal prompts.
- Preserve existing Meta prompt/CTA behavior with focused regression tests.

Validation (2026-09-21): 4 focused Jest suites / 63 tests passed; TypeScript check passed.

Scope: Production consumers remain on the existing Meta path. Connecting validation, prompt assembly and UI to the shared spec remains follow-up work; acceptance of the C1 foundation separately is pending reviewer confirmation.


variations-studio-api/src/platforms/design-checks/googleRsa.ts
```diff
@@ -0,0 +1,32 @@
+import type { PlatformSpec } from '../types';
+
+/**
+ * MED-372/MED-389 design check only, deliberately absent from platformRegistry.
```
quanwangniuniu: lets always avoid add linear issue number in comments and this one is not making any sense here

variations-studio-api/src/platforms/registry.ts
quanwangniuniu, line 1: bad code pattern for the whole file

variations-studio-api/tests/platformCompatibility.test.ts
```diff
@@ -0,0 +1,65 @@
...
+    ['custom user prompt', buildUserPrompt(template, '  Focus on the benefits.  '),
+      'c320fd48d3bd48ae5017af14a32144440a2648c277bf82131121c5a871c3a7b5'],
+    ['default external URL prompt', buildExternalUrlPrompt(pageText, ' \n\t '),
+      'a85ef3150469452a6209c07a3d326d51b7c88a3e52b380fedcf86acbd5c0640c'],
+    ['custom external URL prompt', buildExternalUrlPrompt(pageText, '  Focus on the benefits.  '),
+      '49bdda0b1421428b02e115e614f8874d4405709014d2f8fd011a1b81bbb66093'],
+  ])('preserves the original %s', (_name, prompt, expected) => {
+    expect(sha256(prompt)).toBe(expected);
+  });
+});
```
quanwangniuniu, line 28: what are these mean:
// Byte-for-byte fixtures guard this abstraction-only change against prompt drift.
  it.each([
    ['system prompt', SYSTEM_PROMPT,
      '0a6141ed79be4f188cc3f0f0c2ba7446a6826e3eaa999eb70e6dddd811525d36'],
    ['CTA allowlist', CTA_ENUM_ALLOWLIST,
      'd5e1d1bd79e17cf4e836aec5327d8be44922d0f5c83f3f1d5343756f4c374568'],
    ['default user prompt', buildUserPrompt(template, ' \n\t '),
      '280a29bd112a56a4c2944f4ce77278ed59c9f4be739e841a77569988140b8b7d'],
    ['custom user prompt', buildUserPrompt(template, '  Focus on the benefits.  '),
      'c320fd48d3bd48ae5017af14a32144440a2648c277bf82131121c5a871c3a7b5'],
    ['default external URL prompt', buildExternalUrlPrompt(pageText, ' \n\t '),
      'a85ef3150469452a6209c07a3d326d51b7c88a3e52b380fedcf86acbd5c0640c'],
    ['custom external URL prompt', buildExternalUrlPrompt(pageText, '  Focus on the benefits.  '),
      '49bdda0b1421428b02e115e614f8874d4405709014d2f8fd011a1b81bbb66093'],
  ])('preserves the original %s', (_name, prompt, expected) => {
    expect(sha256(prompt)).toBe(expected);
  });
});

quanwangniuniu (changes requested):

variations-studio-api/src/platforms/registry.ts
PianoMaksim: Fixed, divided into smaller internal functions within the same file, grouped by responsibility.

variations-studio-api/src/platforms/design-checks/googleRsa.ts
PianoMaksim: Fixed, Removed it already

variations-studio-api/tests/platformCompatibility.test.ts
PianoMaksim: These six hash checks verify that adding the platform specs hasn't changed the existing Meta prompt text, CTA allowlist, or prompt-builder output for fixed inputs.
Well, I just realized that, because it doesn't change the existing behavior, thus, maybe I can just delete it on the pr, which means, testing on my local environment and not uploading it. How do you think, remove it or keep it, or use readable expected text so we can see exactly what changed if a test fails? (But it seems it doesn't change)

quanwangniuniu: eveyrthing testing on local stuff should never be pushed to remote

variations-studio-api/src/platforms/registry.ts
quanwangniuniu: still too complex. Actually this should not be that complex

variations-studio-api/tests/platformCompatibility.test.ts
PianoMaksim: Roger that

variations-studio-api/src/platforms/registry.ts
PianoMaksim: I haven't push it sorry, I will push it rn with every issue fixed together

quanwangniuniu (changes requested):
Here are 3 other important problems in PR #828:



1. **The registry is not connected to the production generation or validation path** — `platformRegistry` is created in [`src/platforms/index.ts`](https://github.com/quanwangniuniu/zmarkio/blob/a0c56b41eb1e2ccbc369ad4a4d7119e6201176ac/variations-studio-api/src/platforms/index.ts#L11-L17), but the PR explicitly leaves existing consumers on the old Meta-specific implementation.  
   **Why fix:** The new platform rules can become inconsistent with actual runtime behavior. For example, a field limit or CTA rule defined in `PlatformSpec` may pass registry tests but never be applied when users generate or validate copy. This creates a false sense that platform configuration is centralized when production still has separate logic.

2. **`freeText` CTA policies are not validated** — [`assertCta`](https://github.com/quanwangniuniu/zmarkio/blob/a0c56b41eb1e2ccbc369ad4a4d7119e6201176ac/variations-studio-api/src/platforms/registry.ts#L52-L66) returns after handling `none` and only validates enum CTAs. A `freeText` CTA can reference a missing field, an array field, or an optional field without triggering an error.  
   **Why fix:** Invalid platform specifications can enter the registry and later cause incorrect form rendering, validation, or prompt assembly. Every CTA type should verify that its referenced field exists and is a scalar string field; the tests should include invalid `freeText` configurations.

3. **The exported registry is mutable and can bypass all validation** — [`platformRegistry`](https://github.com/quanwangniuniu/zmarkio/blob/a0c56b41eb1e2ccbc369ad4a4d7119e6201176ac/variations-studio-api/src/platforms/index.ts#L11) is exposed as a mutable `Map`. Code can call `.set()`, `.delete()`, or replace a spec after `createPlatformRegistry` has validated it.  
   **Why fix:** This undermines the registry’s main invariant: every registered platform is validated and unique. A later mutation could introduce duplicate, malformed, or unsupported platform definitions at runtime. Expose a read-only lookup API or freeze/encapsulate the underlying map instead.

PianoMaksim:
1, Meta generation now reads the system prompt, prompt version and slug source from the spec, and applies its CTA policy once when saving — CTA handling has moved out of the provider. The existing Meta prompt text and CTA normalization/fallback behavior are preserved. Copy-length validation, user-prompt builders and UI integration remain follow-up work (C3–C6).

2, Both enum and freeText CTA policies now require a declared, required string field. Missing and array fields were already rejected; I've added the required-field check and tests covering all three invalid cases for both policies.

3, The backing Map is now private. Callers use getPlatformSpec and listPlatformIds instead of the Map. Query results have readonly types, but the spec objects are not cloned.

quanwangniuniu (approved):
Good job

